PR #17, opened 2017-10-10
Make PlayHelper able to handle only a VJoyDevice on player 2

WesselKroos:
I was unable to only use player 2 in the VJoy configuration so I've changed the PlayHelper.py so it doesn't brake anymore. This makes me able to play against my bot in exhibition mode

### DIFF
--- a/PlayHelper.py
+++ b/PlayHelper.py
@@ -5,64 +5,60 @@ import configparser
 class play_helper:
 	
     def __init__(self):
-        self.p1 = pyvjoy.VJoyDevice(1)
-        self.p2 = pyvjoy.VJoyDevice(2)
         config = configparser.RawConfigParser()
         config.read('rlbot.cfg')
         self.p1Enabled = ("True" == config.get('Player Configuration', 'p1Enabled'))
         self.p2Enabled = ("True" == config.get('Player Configuration', 'p2Enabled'))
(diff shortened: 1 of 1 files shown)


tarehart:
Thanks for contributing!

This shouldn't stop us from accepting the pull request, but I want to point out that this conflicts with the own-controller branch, which we apparently never merged in to master.

tarehart:
We opted to merge in own-controller first, so now you're stuck with the conflict resolution, sorry. On the bright side, your change should require less code now, if it's even still necessary at all.

drssoccer55:
Yupp looks like this issue will also happen with the latest changes and needs to be fixed.  Code always assumes 2 vjoy controllers exist which is a bad assumption.

drssoccer55:
@tarehart I believe with your last code change this is taken care of.  We can close this now right?

tarehart:
I think there's still a weird spot here: https://github.com/drssoccer55/RLBot/blob/master/PlayHelper.py#L6

If the user has only one vJoy device configured, and chooses to disable p1 and enable p2 in rlbot.cfg, I imagine this code will break.

I think we can still close this though, since playing against your bot in exhibition mode is pretty easy at this point. (Just have 2 vjoy controllers configured, and disable one of the players in rlbot.cfg)

PR #24, opened 2017-11-23
Added the ability to save game data

dtracers:
This adds the ability to write game data to a file. It will save the input to the bot as an array and the output to the bot as an array.

An official output format does need to be decided.


runner.py
```diff
@@ -67,6 +70,18 @@ def run_agent(terminate_event, callback_event, name, team, index, module_name):
     callbacks = []
     name_dict = dict()
 
+    save_data = True
```
drssoccer55: Read this value from RLBot config

bot_manager.py
```diff
@@ -92,13 +103,23 @@ def run(self):
 
             self.interlocked_exchange_fn(ctypes.byref(player_input_lock), ctypes.c_long(REFRESH_NOT_IN_PROGRESS))
 
+            if self.save_data and game_tick_packet.gameInfo.bRoundActive and old_time is not 0 and not old_time == current_time:
+                self.game_file.writelines(str(game_tick_packet) + '\n')
```
drssoccer55: Need good format for output writing since this is going to be a standard.  Some good discussion in Discord on this.

runner.py
```diff
@@ -67,6 +70,18 @@ def run_agent(terminate_event, callback_event, name, team, index, module_name):
...
+    save_path = os.getcwd() + '\\training'
+    game_name = str(int(round(time.time() * 1000))) + '-' + str(random.randint(0, 1000))
```
drssoccer55: I don't think random part is necessary. Just seconds seems sufficient to me.  Chance of 2 people running at same second is slim.

bot_manager.py
drssoccer55: Talked about writing cstruct bytes to file and reading back to struct with ctypes pack and unpack in the discord as a possibility: https://docs.python.org/3/library/struct.html#struct.unpack

drssoccer55: Note that I mistakenly linked you to struct whereas I am using a ctypes struct.  You can go back and forth from a bytearray to a ctypes struct though.

binary_converter.py
```diff
@@ -0,0 +1,75 @@
+import struct
+import io
+import numpy as np
```
drssoccer55: Just giving a note to myself that numpy is not part of standard python so people will need to install numpy if they don't already have it.  I'm not really against this since numpy is pretty essential but a few people had trouble installing numpy in the past.

drssoccer55:
Closing because this work has been forked to a place where everyone can use numpy.

PR #30, opened 2018-01-19
 Give each game_ticket_packet to the agent exactly once

DomNomNom:
See commit messages for details.

### DIFF
--- a/bot_manager.py
+++ b/bot_manager.py
@@ -72,15 +72,17 @@ class BotManager:
                 game_data_shared_memory.seek(4, os.SEEK_CUR) # Move 4 bytes past error code
                 ctypes.memmove(ctypes.addressof(game_tick_packet), game_data_shared_memory.read(ctypes.sizeof(gd.GameTickPacket)),ctypes.sizeof(gd.GameTickPacket))  # copy shared memory into struct
 
-            # Run the Agent only if the gameInfo has updated.
+            new_module_modification_time = os.stat(agent_module.__file__).st_mtime
+            module_has_changed = new_module_modification_time != last_module_modification_time
+
+            # Run the Agent only if things have updated.
             tick_time = game_tick_packet.gameInfo.TimeSeconds
-            if tick_time != last_tick_time:
+            if tick_time != last_tick_time or module_has_changed:
                 last_tick_time = tick_time
 
                 try:
                     # Reload the Agent if it has been modified.
-                    new_module_modification_time = os.stat(agent_module.__file__).st_mtime
-                    if new_module_modification_time != last_module_modification_time:
+                    if module_has_changed:
                         last_module_modification_time = new_module_modification_time
                         print('Reloading Agent: ' + agent_module.__file__)
                         imp.reload(agent_module)
--- a/bot_manager.py
+++ b/bot_manager.py
@@ -9,10 +9,11 @@ import os
 import rate_limiter
 import sys
 import imp
+import traceback
 
 OUTPUT_SHARED_MEMORY_TAG = 'Local\\RLBotOutput'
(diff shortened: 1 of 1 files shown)


DomNomNom:
I think there is a misunderstanding here:
The rate at which bot scripts will get called while the game is running is still 60 per second due to this being fixed by the producer of the information. You can see in both the Before and After screenshots, it hovers around 0.0166 (1/60).

As for the "iterate while paused" workflow. I could put the check for whether your script has changed on the outside and force one run. I do not see the value of continuously calling the script if the game is paused.

tarehart:
Our dll has no explicit controls on how often it updates the packet, it just does so whenever rocket league happens to execute its playerTick function. That may not be 60Hz for everybody, and it may be subject to change with rocket league updates, etc. We should talk to ccman to get his opinion on how the game probably works and what the risks are.

I'm not keen on letting the dll affect bot execution frequency anyway, I think that's more intuitively a python side responsibility. I wouldn't want dll maintainers to have to worry about the implications of altering the update rate on fairness / bot behavior.

The "iterate while paused" workflow is still broken for people who use grpc, because they do not currently reload their python module at all when making changes.

There's also a "breakpoint while paused" workflow where the dev may step too far and wish to resume execution and break again when the next identical packet comes along.

DomNomNom:
I can fix "iterate while paused" for the usecases you described by having a minimum call rate. Will send a commit soon.


I think it's silly to reject the 99% of the time when this stops Agents from missing tick packets based on the 1% hypothetical scenario of the tick rate changing.
[Don't let the great be the enemy of the good](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good): If you know of a way of setting a maximum of exactly 60hz and not wrongly drop tick packets, you shouldn't block this incremental improvement.
It seems impossible to me to update at exactly 60hz and not miss a packet and not delay input to the game as there is not enough information to make a local decision of whether skipping the current packet is the right choice.

tarehart:
One idea I started thinking about today was having a burst rate that's ~4x faster than normal when an identical packet is detected, continuing for a maximum of one normal frame, or as soon as there is a different packet, whatever happens sooner. That should have the effect of re-aligning the bot loop so that it's querying data shortly after it's available, minimizing round trip latency. Assuming that the game and the bot both tick at ~60 Hz, the alignment should be fairly stable, and the burst behavior would be rarely needed.

Under that system, a bot might do its own detection of identical frames so that it can ignore them and/or return quickly and be ready to receive a fresh packet coming in on the burst.

That gives bots more information and more control, and lets the python side of the framework keep responsibility over bot tick rates.

DomNomNom:
Fell free to write that yourself. My hunch is that it'll still drop packets

tarehart:
Good point, I could imagine packet drops happening if the timing varies due to CPU scheduling or something. Cranking up the sampling frequency to ~61 hz should be adequate, at the expense of requiring bursts more often due to the drift.

I like that your minimum call rate concept simplifies the interface from the bot's perspective. That may be more valuable in practice than giving bots maximum information. What minimum call rate did you have in mind, maybe something like 1 hz?

DomNomNom:
In commit [990bd86](https://github.com/drssoccer55/RLBot/pull/30/commits/990bd867c2024f7b16e1f65be8a8581414d9cb5c) I've chosen 30hz. I'm not too fussed about this value. Happy to change it to something you think is reasonable.

tarehart:
Oh I didn't notice when you added that commit, sorry about that. I'll test this when I get home and probably merge it as-is. 

I'll be experimenting with my 61 Hz + burst idea as a possible refinement in a separate commit. Ping me on discord if you have any more feedback on it.

PR #32, opened 2018-01-30
Added an optional config file that the agent can take in

dtracers:


tarehart:
Giving bots more configuration flexibility is a good idea. I'm worried that this particular design will encourage people to use their cfg files in unpredictable ways, which would hinder our ability to improve the framework in the future without breaking people.

I'm still thinking about this, but how would you feel about having a section in the cfg file called "Bot Params" which could be turned into a dict and handed to the bot's constructor instead of the whole file?

I'm also curious about what your specific use case is.

dtracers:
I can see that working.

For our use we have multiple headers that we grab from to keep the config nice
```
[Bot Location]
# Path to module from runner
# Only need this if RLBot controlled
agent_module = saltie

[Participant Loadout]
name = Saltie10
team_color_id = 27
custom_color_id = 0
car_id = 23
decal_id = 1435
wheels_id = 1728
boost_id = 69
antenna_id = 217
hat_id = 580
paint_finish_1_id = 1978
paint_finish_2_id = 1978
engine_audio_id = 0
trails_id = 1997
goal_explosion_id = 1905

[Model Configuration]
model_package = models.actor_critic.tutorial_model
model_name = TutorialModel
control_scheme = regression_controls
#model_directory = training/data/tutorial_bot_split10-layers/2/trained_variables
teacher = replay_files
num_layers = 4
num_split_layers = 2
num_width = 128
is_graphing = True
is_evaluating = True
exploration_factor = 500
should_graph = False
batch_size = 1
mini_batch_size = 1
```

Here is one of our configs
But in the future we plan to break out model config to be multiple headers so it is clear which parts apply to which model.

do you think we could have a way to exclude participant loadout / bot location so that agents do not have access to those?

tarehart:
I can't think of a reasonable way to exclude sections. Are you sure you can't get by with a single section? You could organize it with newlines and comments: https://docs.python.org/3.4/library/configparser.html#supported-ini-file-structure

You could also store the location of a separate config file and your bot could open and parse it itself.

```
[Bot Location]
# Path to module from runner
# Only need this if RLBot controlled
agent_module = saltie

[Participant Loadout]
name = Saltie10
team_color_id = 27


[Bot Params]

# Model Stuff
model_package = models.actor_critic.tutorial_model
model_name = TutorialModel

# Other stuff
control_scheme = regression_controls

# Supplemental config
external_config_location = my-other-config.cfg
```

dtracers:
I think if push comes to shove we could get by with a single section.
I was considering doing repeated keys.

Though I feel like we should still send a parser in rather than a dict so that we can take advantage of the error checking built in and the type parsing (booleans/float/ints)

dtracers:
Also it looks like we can exclude the other sections by reading the file ourselves and creating a temp in memory file then passing that to the parser.

https://stackoverflow.com/a/25493615/2187510 (different purpose but can achieve the result we want)

tarehart:
Fair point about the convenience of the parser object. What if we just send in the section object that you could get from `configparser['Bot Params']`?

I'm a little sour on the idea of section exclusion because it's unintuitive
 - users would probably be surprised that they can access only their own sections
 - the feature would not be very discoverable. They'd get an empty object passed to their bot, and they'd have to read documentation to figure out how to make it useful. (as opposed to a predefined section which would act as a better signpost)
 - future maintainers may forget to update the blacklist as new framework-level sections are added, and the backwards compatibility issue comes back

dtracers:
I was not aware you can get a section object.
If that is possible then I think that will be the best compromise.

runner.py
```diff
@@ -22,6 +22,7 @@
 BOT_CONFIG_LOADOUT_HEADER = 'Participant Loadout'
 BOT_CONFIG_LOADOUT_ORANGE_HEADER = 'Participant Loadout Orange'
 BOT_CONFIG_MODULE_HEADER = 'Bot Location'
+BOT_CONFIG_AGENT_HEADER = 'Agent Configuration'
```
tarehart: We'd be calling it three different things in the same file, which is confusing - Participant, Bot, and now Agent. I don't think there's any meaningful difference, so I'd suggest calling this "Bot Arguments" or "Bot Parameters"

bot_manager.py
```diff
@@ -21,14 +21,21 @@
...
         self.callbackEvent = callbackEvent
+        self.config_file = config_file
         self.name = name
         self.team = team
         self.index = index
         self.module_name = modulename
 
+    def load_agent(self, agent_module):
+        try:
+            agent = agent_module.Agent(self.name, self.team, self.index, config_file=self.config_file)
```
tarehart: Having the parameter be called config_file seems confusing, because they're really just getting one section. I think it would be more intuitive to give it the same name as the corresponding section header.

Likewise for the BotManager constructor and anywhere else where the variable refers to only one section.

PR #36, opened 2018-02-27
Autorun injector

skyborgff:
autorun injector

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -60,7 +60,7 @@ def main():
     incode=subprocess.call([injector_dir, 'hidden'])
     injector_codes=['INJECTION_SUCCESSFUL','INJECTION_FAILED','MULTIPLE_ROCKET_LEAGUE_PROCESSES_FOUND','RLBOT_DLL_ALREADY_INJECTED','RLBOT_DLL_NOT_FOUND','MULTIPLE_RLBOT_DLL_FILES_FOUND']
     print(injector_codes[incode])
-    if incode!=0:
+    if incode!=0 or incode!=3:
         sys.exit()
 
     # Set up RLBot.cfg
--- a/runner.py
+++ b/runner.py
@@ -53,7 +53,7 @@ def run_agent(terminate_event, callback_event, config_file, name, team, index, m
                                 index, module_name)
     bm.run()
     
-def injectDLL()
+def injectDLL():
     """
     Calling thid function will inject the DLL with as 'hidden'
     DLL will return status codes from 0 to 5 wich correspond to injector_codes
--- a/runner.py
+++ b/runner.py
@@ -53,7 +53,14 @@ def run_agent(terminate_event, callback_event, config_file, name, team, index, m
                                 index, module_name)
     bm.run()
     
-def inject()
+def injectDLL()
+    """
+    Calling thid function will inject the DLL with as 'hidden'
+    DLL will return status codes from 0 to 5 wich correspond to injector_codes
+    DLL injection only valid if codes are 0->'INJECTION_SUCCESSFUL' or 3->'RLBOT_DLL_ALREADY_INJECTED'
+    It print the output code and if it's not valid it will kill the runner.py
+    If RL isn't running the Injector will stay hidden waiting for RL to open and inject as soon as it does
+    """
     # Inject DLL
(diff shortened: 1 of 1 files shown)


runner.py
```diff
@@ -54,6 +55,12 @@ def run_agent(terminate_event, callback_event, config_file, name, team, index, m
 
 
 def main():
+    # Inject DLL
+    injector_dir = os.path.join(os.path.dirname(__file__), "RLBot_Injector.exe")
+    incode=subprocess.call([injector_dir, 'hidden'])
+    injector_codes=['INJECTION_SUCCESSFUL','INJECTION_FAILED','MULTIPLE_ROCKET_LEAGUE_PROCESSES_FOUND','RLBOT_DLL_ALREADY_INJECTED','RLBOT_DLL_NOT_FOUND','MULTIPLE_RLBOT_DLL_FILES_FOUND']
```
drssoccer55: We should fail on some of these codes.

skyborgff: what do you mean?

drssoccer55: INJECTION_SUCCESSFUL or RLBOT_DLL_ALREADY_INJECTED seem like success statuses. The others should probably terminate runner. Also this code will likely change with the GUI in v4 branch so might be better to implement as part of v4.

skyborgff: Fixed (i believe)

runner.py
```diff
@@ -54,6 +55,14 @@ def run_agent(terminate_event, callback_event, config_file, name, team, index, m
...
+    print(injector_codes[incode])
+    if incode!=0 or incode!=3:
```
drssoccer55: If statement will always be true.
incode = 0 would be -> False or True -> True -> exit()

It would be nicer to compare to the strings instead of index too. Ex. not equal to INJECTION_SUCCESSFUL and not equal to RLBOT_DLL_ALREADY_INJECTED.

tarehart: I'm pretty sure this is equivalent to `if true`. You probably want `if not (incode == 0 or incode == 3)`

skyborgff: yep

runner.py
```diff
@@ -54,6 +55,14 @@ def run_agent(terminate_event, callback_event, config_file, name, team, index, m
 
 
 def main():
+    # Inject DLL
```
dtracers: can we actually move it to its own method that then gets called from main?

That way we can have more logic around if we want to perform dll injection at all

skyborgff:
i moved it to it's own method now

runner.py
```diff
@@ -52,17 +52,25 @@ def run_agent(terminate_event, callback_event, config_file, name, team, index, m
...
     injector_dir = os.path.join(os.path.dirname(__file__), "RLBot_Injector.exe")
     incode=subprocess.call([injector_dir, 'hidden'])
     injector_codes=['INJECTION_SUCCESSFUL','INJECTION_FAILED','MULTIPLE_ROCKET_LEAGUE_PROCESSES_FOUND','RLBOT_DLL_ALREADY_INJECTED','RLBOT_DLL_NOT_FOUND','MULTIPLE_RLBOT_DLL_FILES_FOUND']
-    print(injector_codes[incode])
-    if incode!=0 or incode!=3:
+    injection_status=injector_codes[incode]
+    print(injection_status)
+    
+    if injection_status=='INJECTION_SUCCESSFUL' or injection_status=='RLBOT_DLL_ALREADY_INJECTED':
+        return injection_status
```
Sciguymjm: Just a nit-pick but you can do `injection_status in [<valid statuses>]` (list of valid statuses) instead of comparing to each valid one.

skyborgff: done

runner.py
```diff
@@ -51,9 +52,26 @@ def run_agent(terminate_event, callback_event, config_file, name, team, index, m
     bm = bot_manager.BotManager(terminate_event, callback_event, config_file, name, team,
                                 index, module_name)
     bm.run()
-
+    
+def inject()
```
dtracers: can we name this to
injectDLL

And add a comment in the format of 
"""
Comment here and says what it does and the situations under which it kills the app
"""

drssoccer55: ooo raising code standards on documentation :)

PR #38, opened 2018-03-01
Granting bot processes high priority, and isolating teams on separate cpus

tarehart:
This will hopefully prevent CPU hogging bots from affecting their opponents.
The process priority increase should help everybody.

This includes a breaking change to grpc_client.py: You now need to call make_grpc_agent with address and port as separate params. I don't think many people are using it yet, so I'm willing to dry any and all tears.

This also requires the installation of psutil via `pip install psutil`, so we would have to update the setup instructions.

### DIFF
--- a/grpcsupport/grpc_client.py
+++ b/grpcsupport/grpc_client.py
@@ -7,7 +7,7 @@ from .protobuf import game_data_pb2_grpc
 
 
 def make_grpc_agent(address, port):
-    '''
+    """
         parameters:
             address - address for grpc connection. example: `localhost`
             port - port for grpc connection. example: 34865
@@ -16,7 +16,7 @@ def make_grpc_agent(address, port):
             grpc server and returns its response.
 
         For an example use of this, see: agents/java_demo
-    '''
+    """
     server_address = address + ':' + str(port)
 
     class GrpcForwardingAgent:
--- a/grpcsupport/grpc_client.py
+++ b/grpcsupport/grpc_client.py
@@ -1,19 +1,23 @@
 import grpc
 import time
+import msvcrt
+import psutil
 from . import proto_converter
 from .protobuf import game_data_pb2_grpc
 
 
-def make_grpc_agent(server_address):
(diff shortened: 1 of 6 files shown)


runner.py
```diff
@@ -207,5 +223,48 @@ def main():
...
+        for team, team_set in team_pids_map.items():
+            team_set -= shared_pids
+
+        for team, team_pids in team_pids_map.items():
+            team_cpu_offset = cpus_per_team * team
+            team_cpus = list(range(cpu_count - cpus_per_team - team_cpu_offset, cpu_count - team_cpu_offset))
+            for pid in team_pids:
+                p = psutil.Process(pid)
+                p.cpu_affinity(team_cpus)  # Restrict the process to run on the cpus assigned to the team
+                p.nice(psutil.HIGH_PRIORITY_CLASS)  # Allow the process to run at high priority
```
drssoccer55: Wonder if this could lead to more choppy recordings since game capture can be cpu intensive.

tarehart: Yep, that's possible. I'll do some testing with both recording and streaming.

grpcsupport/grpc_client.py
```diff
@@ -29,6 +33,17 @@ def __init__(self, name, team, index):
                 print("Exception when trying to connect to grpc server: " + str(e))
                 pass
 
+        def get_extra_pids(self):
```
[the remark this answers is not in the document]
tarehart: I tinkered with this a little, but the multiprocessing situation is very complicated.

However, blocking is probably fine: If an agent is taking a long time to get process ids, that's probably because it hasn't started up yet, so it's screwed for the first kickoff anyhow. It won't block the other agent from running.

I'm inclined to leave this as-is.

runner.py
tarehart: Here's some stress testing, I don't think anything got worse: https://www.twitch.tv/videos/234291596?t=00h07m25s

Streaming at 720p which is what I ended up with during the tournament after tinkering.

PR #45, opened 2018-03-17
Added Wheels to item parser

skyborgff:


Sciguymjm:
I feel like this entire script could be more concise since everything but skins are identical in function. Think about refactoring so that you don't reuse code.

Otherwise functionally it looks fine.

skyborgff:
@Sciguymjm i wont be home more today, but tomorrow i'll look into it :)

RLBotFramework/utils/item_id_parser.py
```diff
@@ -1,7 +1,7 @@
 # data in the form of: ID,FullName,AssetPackageName,AsciiLabel,PackID,Quality,UnlockMethod
 
 dump_file = input('Data dump (input) file name?->')
-data = open(dump_file, 'r', encoding='utf-8-sig')
```
tarehart: Did you get rid of this encoding param on purpose?

skyborgff: Yes i did, the first time i used it it needed that( i think i downloaded the txt).
This time, i copied the text and created the txt myself, and it gave me an error because it was not encoded in utf-8-sig.

tarehart:
I decided to go ahead and merge this. Sciguymjm has a valid comment, but I think that would be good for a future pull request. Thanks for fixing this!

PR #46, opened 2018-03-18
 Cleaned up file, added automatically date, and fixed @Sciguymjm request

skyborgff:
@Sciguymjm i did the refactoring :)


RLBotFramework/utils/item_id_parser.py
```diff
@@ -28,15 +34,14 @@
...
-    if (ID=='0' or AssetPackageName == 'seasonlogos' or
-    AssetPackageName == 'itemcontainer' or
-    AssetPackageName == 'bots' or
-    'key' in AssetPackageName or
-    'pack' in AssetPackageName):
+    if (ID=='0' or AssetPackageName == 'seasonlogos' or AssetPackageName == 'itemcontainer' or
+    AssetPackageName == 'bots' or 'key' in AssetPackageName or 'pack' in AssetPackageName):
         continue
+
     elif 'body' in FullName:
```
dtracers: Couldn't you make these all into using a dictionary?
Instead of doing the same thing?

skyborgff: I actually thought of it, but I thought it would be harder to update.

RLBotFramework/utils/item_id_parser.py
```diff
@@ -45,95 +50,64 @@
...
-    clean_file.write('\n')
-
-clean_file.write('```\n\n### Engine_Audio\n```\n')
-for engineaudio in Engine_Audio:
-    clean_file.write(engineaudio)
-    clean_file.write('\n')
-clean_file.write('```')
-
-clean_file.close()
+def write_file(types):
```
Sciguymjm: Definitely much better for writing the file out.

RLBotFramework/utils/item_id_parser.py
Sciguymjm: I agree with dtracers in that you could make a dictionary that contains all of this. Instead of individual arrays it could be `the_dict['Decals']` or `the_dict['Antennas']` so that you could just add a line later that adds a new type. Then the code below would automatically iterate over all the keys in the dictionary.

skyborgff:
@Sciguymjm i implemented a dictionary, now it is more clean

PR #49, opened 2018-03-22
Change config option to participant_type

TheBlocks:
The rlbot.cfg file now handles the types of participants like this:
```
participant_type_0 = human
participant_type_1 = rlbot
participant_type_2 = psyonix
participant_type_3 = possessed_human
```

### DIFF
--- a/RLBotFramework/gui/agent_frames/agent_frame.py
+++ b/RLBotFramework/gui/agent_frames/agent_frame.py
@@ -25,6 +25,11 @@ class AgentFrame(BaseAgentFrame):
         self.player_type = tk.StringVar(value="RLBot")
         self.agent_path = tk.StringVar()
         self.latest_looks_path = tk.StringVar()
+        self.player_type_config_gui = {"Human": "human",
+                                       "Psyonix Bot": "psyonix",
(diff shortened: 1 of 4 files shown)


TheBlocks:
Fixed typo

TheBlocks:
Need to add same functionality to GUI runner

TheBlocks:
Fixed for runner_gui.py

RLBotFramework/gui/agent_frames/agent_frame.py
```diff
@@ -68,17 +67,17 @@ def refresh_widgets(self):
         for widget in self.grid_slaves(row=3, column=1):
             widget.grid_forget()
         if self.player_type.get() == "Human":
-            self.is_bot.set(False)
-            self.rlbot_controlled.set(False)
+            self.participant_type.set("human")
```
dtracers: this human and the Human above should be able to use the exact same value with a constant declared at the top of the file.

The same should go for the other bot types

TheBlocks:
Updated to fit with requested change. Now uses a dict for lookup instead of if-else statements.

PR #50, opened 2018-03-23
Load bots from anywhere in the file system

tarehart:


RLBotFramework/utils/class_importer.py
```diff
@@ -1,75 +1,73 @@
...
-        # grabs only the first one
-        return agent
-    except ModuleNotFoundError as e:
-        log_warn('ModuleNotFoundError: %s\nsub module %s not found using %s instead', [str(e), str(module_name),
-                                                                                       str(base_class)])
-    except Exception as e:
-        log_warn('Error: %s\n%s not found using %s instead', [str(e), str(module_name), str(base_class)])
+    init_name = os.path.join(dir_name, "__init__.py")
+    if not os.path.exists(init_name):
+        with open(init_name, 'w'):  # Create an empty __init__.py file if it doesn't exist
```
dtracers: you do not need the with
you should be able to just do
`open(init_name, 'w')`

RLBotFramework/utils/class_importer.py
```diff
@@ -1,75 +1,73 @@
...
-    super_classes = module.__bases__
-    for super_class in super_classes:
-        if super_class == base_class:
-            return True
-    return False
+class AgentLoadData:
+    def __init__(self, module_spec, base_class):
+        self.module_spec = module_spec
+        self.base_class = base_class
+        self.agent_class = None
```
dtracers: I think there should be a method to get the agent class instead of directly accessing the agent class
In case we want to perform some safe logic around the agent_class being None

RLBotFramework/utils/class_importer.py
tarehart: I believe `with` has the advantage of closing the resource when the block exits.

RLBotFramework/utils/class_importer.py
tarehart: Yep, could be useful. Added it, but didn't do any logic for None right now because currently it is set in the constructor during reload().

RLBotFramework/utils/class_importer.py
```diff
@@ -1,75 +1,74 @@
...
-        if super_class == base_class:
-            return True
-    return False
+class AgentLoadData:
+    def __init__(self, python_file, base_class):
+        self.python_file = python_file
+        self.base_class = base_class
+        self.agent_class = None
+        self.agent_module = None
+        self.reload()
```
dtracers: are we sure we want to call reload in the constructor?
It could potentially cause issues.

Also how does this work with the importing of extensions

RLBotFramework/utils/class_importer.py
```diff
@@ -1,75 +1,74 @@
...
-from RLBotFramework.utils.logging_utils import log_warn
 
 
-def is_extends_base_class(module, base_class):
-    super_classes = module.__bases__
-    for super_class in super_classes:
-        if super_class == base_class:
-            return True
-    return False
+class AgentLoadData:
```
dtracers: it looks like this applies to agents and extensions.
Should we come up with a name that applies to both?

RLBotFramework/utils/class_importer.py
tarehart: Doesn't seem particularly dangerous to me. Is there something especially likely or especially bad about constructor crashes?

RLBotFramework/utils/class_importer.py
tarehart: Sure, I'll rename it.

RLBotFramework/utils/class_importer.py
Sciguymjm: I'm pretty sure the object doesn't get created.

tarehart: I think that in the old days, exceptions coming from a constructor were harder to handle properly or harder to debug, so people started giving advice about making constructors do very little. I don't have any reason to believe that such advice is relevant in Python.

dtracers:
For me it's less that and if we want to handle this exception and not have
it crash everything.

If we handle it then we need a case for still getting the object / setting
it to none.

If the error is thrown outside then only that one call needs to be wrapped
and handled.

tarehart:
I don't think that an AgentLoadData instance with fields that are None is a good outcome. Much more pleasant to give a guarantee that if you have it, it's fully initialized and operational.

In other words, the reasonable exception handling for this does not require an AgentLoadData object.

RLBotFramework/utils/class_importer.py
Sciguymjm: I don't see there being an actual issue with this, I'm just answering the question. It looks fine.

PR #53, opened 2018-03-25
Dll update

ccman32:


tarehart:
I tested this briefly and it works for me!

PR #54, opened 2018-03-26
Adding framework support for java bots that communicate directly with dll.

tarehart:
Later, we can use gradle to compile this into a jar, and include the jar in a maven publication. Java bots will be able to pull in this code as a dependency. Easy for them, and more control for us in terms of fairness and vending updates.

I've successfully tested this in-game.

### DIFF
--- a/src/main/java/rlbot/Bot.java
+++ b/src/main/java/rlbot/Bot.java
@@ -1,0 +1,12 @@
+package rlbot;
+
+import rlbot.api.GameData;
+
+public interface Bot {
+
+    int getIndex();
+
+    GameData.ControllerState processInput(GameData.GameTickPacket request);
+
+    void retire();
+}
--- a/src/main/java/rlbot/manager/BotManager.java
+++ b/src/main/java/rlbot/manager/BotManager.java
@@ -1,0 +1,86 @@
+package rlbot.manager;
+
+import com.google.protobuf.InvalidProtocolBufferException;
+import rlbot.Bot;
+import rlbot.api.GameData;
+import rlbot.cpp.RLBotDll;
+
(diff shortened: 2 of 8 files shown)


src/main/java/rlbot/manager/BotManager.java
```diff
@@ -0,0 +1,86 @@
...
+
+        botProcesses.computeIfAbsent(index, (idx) -> {
+            Thread botThread = new Thread(() -> doRunBot(bot));
+            botThread.start();
+            return new BotProcess(bot, botThread);
+        });
+    }
+
+    private void doRunBot(Bot bot) {
+        while (keepRunning) {
```
dtracers: instead of a while loop you could try using a timer.
It would allow us to call it more with the 120 hz like we do with the python bots,
I think there is a way to have the timer wait until the previous call has completed.

src/main/java/rlbot/Bot.java
```diff
@@ -0,0 +1,12 @@
+package rlbot;
+
+import rlbot.api.GameData;
+
+public interface Bot {
```
dtracers: java docs

src/main/java/rlbot/manager/BotManager.java
tarehart: This loop does not need a timer. `dinnerBell.wait(1000);` will cause this thread to wait until a fresh packet is available.

Your comment applies better to the doLoop method where I'm sleeping for 16 ms.  There's room for improvement there, and one solution is 120hz, but I'd like to hold off on that for now because I think we can do better. We could make the dll block until a fresh packet is available, or we could implement your long-lasting callback idea.

PR #55, opened 2018-03-29
Restoring the ability to shut down RLBot gracefully. Benching protoBot.

tarehart:


tarehart:
This also improves performance significantly, for me at least.

RLBotFramework/setup_manager.py
```diff
@@ -113,9 +117,26 @@ def run(self):
             except queue.Empty:
                 pass
             except Exception as ex:
-                print(ex)
+                self.logger.error(ex)
                 pass
 
+        self.logger.info("Shutting Down")
```
dtracers: can we actually move this shutdown code to its own place in runner?

Mainly I think we should have a way to dynamically kill bots without killing the main thread.

RLBotFramework/utils/structures/quick_chats.py
```diff
@@ -119,27 +120,30 @@ def create_queue_for_bot(self, index, team):
...
-                if bots[0] != team and team_only:
-                    # do not send to other team if team only
-                    continue
-                bots[1].put((index, team, message_details))
-            self.game_interface.send_chat(index, team_only, message_details)
-
-    def start_manager(self):
-        thread = Thread(target=self.process_queue, args=())
+    def process_queue(self, quit_event, quit_callback):
+        while self.game_running and not quit_event.is_set():
```
dtracers: why leave the game running check here if we are using the quit event check?

tarehart: Good point, I'll remove it.

RLBotFramework/setup_manager.py
tarehart: Done, and also removed a few things that proved unnecessary

PR #56, opened 2018-04-01
Moved Match settings into its own file

dtracers:
Added match settings into its own file.

This will break the gui (but it currently does not work)
And we are moving to a new gui implementation.
This also adds match settings and mutator settings.
In addition this also changes to using the bot type instead of the two booleans.
```
participant_type_0 = human
participant_type_1 = rlbot
participant_type_2 = psyonix
participant_type_3 = possessed_human
```


RLBotFramework/parsing/agent_config_parser.py
```diff
@@ -0,0 +1,176 @@
...
+        is_rlbot = False
+    elif bot_type == 'rlbot':
+        is_bot = True
+        is_rlbot = True
+    elif bot_type == 'psyonix':
+        is_bot = True
+        is_rlbot = False
+    elif bot_type == 'possessed_human':
+        # this is an rlbot but a very specific rlbot
+        is_bot = True
```
tarehart: is_bot = False

dtracers: is_bot will still be true.
Because it is running an agent (notice the comment right above that line :P)

tarehart: I feel that training_bot is a bad name, and also that possessed_human is a bad name for what you're using it for.

I originally came up with "possessed_human" as a short way of saying "a 'human' player from the game's perspective which is actually controlled by rlbot". Obviously possessed_human was not an ideal name because it got lost in translation when you wrote this code. I'd say it's an even worse name for "a bot that drives based on input from a controller".

I'm not a fan of training_bot because it could (rarely) be used outside of training mode, and generally I don't like mixing up what something is with how it's expected to be used.

PR #57, opened 2018-04-03
Setting up gradle building and publishing

tarehart:
Most of this diff is me moving all the python files into a folder so we can package them into a zip file in a predictable way.

Capabilities:
- `./gradlew build` will do the following:
  - install the python dependencies in requirements.txt
  - take the protobuf python file from our RLBotProtoMaven dependency and put it where runner can see it
  - compile the java framework into a jar
- `./gradlew publishToMavenLocal` will allow us to consume this package from elsewhere via jitpack, and it creates these artifacts:
  - -python.zip: includes all our python files, plus the dlls and injector
  - .jar: bot manager for java bots
  - -sources.jar: source files for the above
  - .pom: maven metadata
  

This enables us to later make a sample bot repository / zip file which does not include the framework code. The instructions would be something like this:
- Clone (or just download) our sample bot repository.
- Double click runner.bat (hypothetical)
  - It executes `./gradlew build`
    - causing the framework files to be downloaded
    - and all the python dependencies in the framework's requirements.txt to be installed
  - and then executes `python runner_GUI.py` to pop open our GUI

Depending on how the build.gradle file is written in the sample bot repository, this could cause the framework to auto-update every time runner.bat is executed.


build.gradle
```diff
@@ -0,0 +1,97 @@
...
+
+processResources.dependsOn pipInstallRequirements
+processResources.dependsOn assembleProto
+
+task sourcesJar(type: Jar, dependsOn: classes) {
+    classifier = 'sources'
+    from sourceSets.main.allSource
+}
+
+def gitignoreToList(File f) {
```
dtracers: What is this used for?

Can we add a comment of it's purpose?

dtracers: I think more generally add comments about what each section independently does.

PR #59, opened 2018-04-08
Modify sys.path per bot process

RLMarvin:


tarehart:
Would you mind pulling `str(module_name).replace(".", "\\") + ".py"` out into a function and putting a comment on it? For example, "Converting a module name to a file path, e.g. mybot.util -> mybot\util.py. We're doing it this way to be backwards-compatible with existing bot configs."

PR #61, opened 2018-04-09
Don't run pip during build. Removing useless line in gitignore.

tarehart:
Running pip during the build is unnecessary, plus jitpack can't handle it:
https://jitpack.io/com/github/RLBot/RLBot/0.0.1/build.log

The pipInstallRequirements task can still be used directly if desired.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,8 +116,6 @@ ENV/
 # Gradle
 .gradle
 
-apikey
-
 # Generated integration test data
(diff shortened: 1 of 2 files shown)


tarehart:
Proof that it works better with jitpack now: https://jitpack.io/com/github/tarehart/RLBot/gradle-build-9a3fb5200e-1/build.log

PR #72, opened 2018-04-30
 Updating protos. Using capnproto and boost shared memory to get game packet.

tarehart:
Lots of serious changes in here.

Heavy changes to the proto specs:
- Boost pad locations are separated out and put in the field info packet.
- Location, rotation, velocity, and angular velocity have been placed inside a "physics" message
- Various naming improvements
- Experimental 'update game state' messages

Game data packet and field info are now using boost shared memory to get from the core dll to the interface dll. Their format in shared memory is capnproto.

Quality of life enhancements for developers:
- `copyDlls` gradle task
- Build task does all protobuf and capnproto file generation, and sends files where they belong.

When this is merged, I will push an update to rlbot-core, RLBotPythonExample, and RLBotJavaExample.


src/main/proto/game_data.proto
```diff
@@ -0,0 +1,158 @@
...
+  optional float pitch = 3; // -1 for nose down, 1 for nose up
+  optional float yaw = 4; // -1 for full left, 1 for full right
+  optional float roll = 5; // -1 for roll left, 1 for roll right
+  optional bool jump = 6;
+  optional bool boost = 7;
+  optional bool handbrake = 8;
+}
+
+message PlayerInput {
+  required int32 player_index = 1;
```
dtracers: proto fields should not be required.
Especially if we ever plan on writing them to a file or making backwards compatible changes in the future

src/main/cpp/RLBotInterface/RLBotInterface/GameFunctions/GameFunctions.hpp
```diff
@@ -15,7 +15,7 @@ extern "C"
 	namespace GameFunctions
 	{
 		DLL_EXPORT RLBotCoreStatus RLBOT_CORE_API UpdateLiveDataPacket(LiveDataPacket* pLiveData);
-		DLL_EXPORT RLBotCoreStatus RLBOT_CORE_API UpdateMatchDataPacket(MatchDataPacket* pMatchData);
+		// DLL_EXPORT RLBotCoreStatus RLBOT_CORE_API UpdateMatchDataPacket(MatchDataPacket* pMatchData);
```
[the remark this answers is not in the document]
tarehart: Because it was never functional, and commenting it out allowed me to get rid of the MatchDataPacket struct. My intent is to use boost queues for setting game state.

src/main/proto/game_data.proto
tarehart: I'm reluctant to make everything optional because I feel like we'd need to do an unreasonable number of checks in c++ to avoid crashes.

src/main/cpp/RLBotInterface/RLBotInterface/GameFunctions/GameFunctions.hpp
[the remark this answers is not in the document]
dtracers: so this is a temporary commented out function?

What method will external clients be calling to update game state?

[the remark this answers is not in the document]
dtracers: okay sg

src/main/proto/game_data.proto
dtracers: we can perform those checks client side.  but making it optional is safer on the wire.  if we remove a field at a later point then everything written to disk with that message (or between dll versions) become incompatible

tarehart: OK, fair enough. Does that advice apply to very primitive things like Vector3 and Rotator, or do you like those the way they are with the *Partial counterparts?

dtracers: generally it is supposed to apply to everything.
But I really doubt we would change the makeup of Vector3 or Rotator so I am fine with leaving those as required.

PR #77, opened 2018-05-12
Moved some CPP files to make them clearer

dtracers:


src/main/cpp/RLBotInterface/RLBotInterface/RLBotInterface.vcxproj
```diff
@@ -1,248 +1,254 @@
...
-  </ImportGroup>
-  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
-    <PropertyGroup>
-      <ErrorText>This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
-    </PropertyGroup>
-    <Error Condition="!Exists('..\packages\protobuf.2.6.1.1\build\native\protobuf.targets')" Text="$([System.String]::Format('$(ErrorText)', '..\packages\protobuf.2.6.1.1\build\native\protobuf.targets'))" />
-    <Error Condition="!Exists('..\packages\protobuf-vc140.2.6.1.1\build\native\protobuf-vc140.targets')" Text="$([System.String]::Format('$(ErrorText)', '..\packages\protobuf-vc140.2.6.1.1\build\native\protobuf-vc140.targets'))" />
-    <Error Condition="!Exists('..\packages\boost.1.67.0.0\build\boost.targets')" Text="$([System.String]::Format('$(ErrorText)', '..\packages\boost.1.67.0.0\build\boost.targets'))" />
-  </Target>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
dtracers: I am using the default settings of git and the default settings of VS

src/main/cpp/RLBotInterface/RLBotInterface/BoostUtilities/BoostUtilities.hpp
```diff
@@ -0,0 +1,52 @@
...
+This typedef is advice from one of the boost maintainers on how to make message queues work between 32 and 64 bit processes.
+It looks pretty janky. I believe the reason it's not "fixed" in the library is that it can't be done consistently on windows vs linux.
+
+"In general I regret putting message_queue in Interprocess, as IMHO it's not good enough to be in the library. Some people find it useful, though."
+https://lists.boost.org/Archives/boost/2014/06/214746.php
+*/
+typedef boost::interprocess::message_queue_t< boost::interprocess::offset_ptr<void, boost::int32_t, boost::uint64_t>> interop_message_queue;
+
+
+#define BEGIN_GAME_FUNCTION(structName, name)	GameInput* pGameInput = FileMappings::GetGameInput(); \
```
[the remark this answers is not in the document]
dtracers: These lines are going away as we move everything to boost.  I would rather them be in one place and have one change needed then have it be in a different place and have to change multiple files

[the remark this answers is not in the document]
tarehart: Most people would assume that this code is boost specific and would not have the confidence to delete it.

Maybe you could create a separate header file for deprecated stuff like this and document it as such.

src/main/cpp/RLBotInterface/RLBotInterface/GameFunctions/GameFunctions.cpp
```diff
@@ -8,218 +8,19 @@
...
-		// The lock will be released when this object goes out of scope
-		boost::interprocess::sharable_lock<boost::interprocess::named_sharable_mutex> myLock(*mtx);
-
-		boost::interprocess::offset_t size;
-		shm->get_size(size);
-		if (size == 0)
-		{
-			// Bail out early because mapped_region will freak out if size is zero.
-			ByteBuffer empty;
-			empty.ptr = new char[1]; // Arbitrary valid pointer to an array. We'll be calling delete[] on this later.
```
ccman32: Is this allocation really necessary?

src/main/cpp/RLBotInterface/RLBotMessages/BoostUtilities/BoostUtilities.hpp
```diff
@@ -0,0 +1,53 @@
+#ifndef BOOSTUTILITIES_HPP
+#define BOOSTUTILITIES_HPP
+
+#include "../Messages.hpp"
+
+#include <boost\interprocess\ipc\message_queue.hpp>
+#include <boost/interprocess/shared_memory_object.hpp>
```
ccman32: Only backslashes should be used here.

src/main/cpp/RLBotInterface/RLBotMessages/BoostUtilities/BoostUtilities.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include <boost\interprocess\ipc\message_queue.hpp>
+#include <boost/interprocess/shared_memory_object.hpp>
+#include <boost/interprocess/mapped_region.hpp>
+#include <boost\interprocess\sync\named_sharable_mutex.hpp>
+#include <boost\interprocess\sync\sharable_lock.hpp>
+
+#include "BoostConstants.hpp"
+
+// Adds the proto to the boost queue with message size checking and buffer size checking
+#define ADD_TO_BOOST_QUEUE(queue, proto, protosize, maxsize)	if (protosize >= maxsize) { \
```
ccman32: This is ok for now but I don't think it's a good long-term solution. Maybe it's better to put this code in a new function and call that function from wherever it needs to be called.

src/main/cpp/RLBotInterface/RLBotInterface/GameFunctions/GameFunctions.cpp
tarehart: This is my code, and I did this because otherwise `delete []` throws an error. There might be a better way to do it.

src/main/cpp/RLBotInterface/RLBotMessages/BoostUtilities/BoostUtilities.hpp
tarehart: My code again, but you're right.

src/main/cpp/RLBotInterface/RLBotMessages/BoostUtilities/BoostUtilities.hpp
tarehart: I agree. See https://github.com/RLBot/RLBot/commit/70d733ea8344d6bb971533a5e14c3fa3fe1d75fc for an alternative technique.

src/main/cpp/RLBotInterface/RLBotInterface/GameFunctions/GameFunctions.cpp
ccman32: Where is that delete call you are mentioning?

tarehart: There are a couple examples of it visible in the diff. Check here if it's easier: https://github.com/RLBot/RLBot/blob/v4/src/main/cpp/RLBotInterface/RLBotInterface/GameFunctions/GameFunctions.cpp#L204

ccman32: Well I'd rather set it to a nullptr and only call delete when it's not a nullptr. However if it's easier we can leave it as it is for now.

tarehart: Ah, I didn't think of that. I can fix that later.

PR #84, opened 2018-05-19
Team Color ID and Exception fix

RLMarvin:
Expanding valid team color id range and bringing back raised exceptions


rlbot.cfg
```diff
@@ -35,13 +35,13 @@ participant_team_7 = 1
...
 
-# Accepted values are "human", "rlbot", "psyonix", and "part_member_bot"
+# Accepted values are "human", "rlbot", "psyonix", and "party_member_bot"
 # You can have up to 4 local players and they must be activated in game or it will crash.
 # If no player is specified you will be spawned in as spectator!
 # human - not controlled by the framework
 # rlbot - controlled by the framework
 # psyonix - default bots (skill level can be changed with participant_bot_skill
-# part_member_bot - controlled by the framework but the game detects it as a human
+# party_member_bot - controlled by the framework but the game detects it as a human
```
dtracers: these need to be changed in the config creation because they are auto generated

RLMarvin: seems like they already correct in the config parser, the typo seems to only exist in rlbot.cfg

PR #90, opened 2018-05-22
Removing the dependency on Java for basic development in other languages

tarehart:


copy-dlls.bat
```diff
@@ -0,0 +1,30 @@
...
+    echo Cannot copy core dll because more than one core directory was found. Please move one of them.
+    EXIT
+)
+
+@echo on
+
+copy /Y "%core_folder%\RLBot Core\Bin\Win32\Debug-SDK\RLBot Core.dll" ^
+ .\src\main\python\RLBotFramework\dll\RLBot_Core.dll
+
+
```
DomNomNom: Trailing whitespace

PR #95, opened 2018-05-28
Allowing bots to request a shareable helper process.

tarehart:
This is needed for Scratch bots. Multiple scratch bots will want to share a single websocket connection. Might be useful for other stuff too.


src/main/python/rlbot/botmanager/agent_metadata.py
```diff
@@ -0,0 +1,11 @@
+from rlbot.botmanager.helper_process_request import HelperProcessRequest
+
+
+class AgentMetadata:
+    def __init__(self, index: int, name: str, team: int, pids: set,
```
gyfis: With the number of positional arguments it may be good to consider using `**kwargs` instead.

src/main/python/rlbot/setup_manager.py
```diff
@@ -120,14 +125,38 @@ def run(self):
...
+        create the process. Once the process is running, feed the agent metadata to it.
+
+        If a process is created here, the pid will be added to the agent metadata.
+        """
+
+        helper_req = agent_metadata.helper_process_request
+
+        if helper_req is not None:
+            if helper_req.key not in self.helper_process_map:
+                metadata_queue = mp.Queue()
```
gyfis: I'd consider moving the `metadata_queue` creation into a separate method, e.g. like

```python
if helper_req.key not in self.helper_process_map:
    self.helper_process_map[helper_req.key] = SetupManager.create_metadata_queue(agent_metadata, self.quit_event)
```

And with that, I'd also consider moving `run_helper_process` and `create_metadata_queue` into a separate file, for example `helper_utils.py`

src/main/python/rlbot/botmanager/agent_metadata.py
tarehart: I prefer it like this so the IDE can infer types.

src/main/python/rlbot/setup_manager.py
tarehart: Refactored into a separate file.

src/main/python/rlbot/botmanager/agent_metadata.py
gyfis: There is also the option of using required keyword arguments (which I believe would work with the types as well): https://stackoverflow.com/a/36446353/1581558

```python
def func2(a, *, b):
    ...
```

This is something I didn't know of before, and was apparently defined for Python 3.x in [PEP 3102](https://www.python.org/dev/peps/pep-3102/)

tarehart: I don't think that would improve readability. The current function signature already supports calls like `AgentMetadata(index=self.index, name=self.name, ...` but I didn't bother because the variable names are already obvious.

I don't think we need any of the flexibility that kwargs offers.

PR #115, opened 2018-05-31
Improved some docs + minor refactor

gyfis:


src/main/python/rlbot/botmanager/agent_metadata.py
```diff
@@ -2,8 +2,20 @@
...
+    """
+    This class is used for storing and passing around the information about a single agent.
+    TODO: Currently overlaps with the information that `BotManager` has.
+    """
+
     def __init__(self, index: int, name: str, team: int, pids: set,
                  helper_process_request: HelperProcessRequest = None):
+        """
+        :param index: The player index, i.e. "this is player number <index>". Will be passed to the bot's constructor.
+        :param name: name which will be passed to the bot's constructor. Will probably be displayed in-game.
```
dtracers: This is the name displayed in game but it is important to note (for you maybe not in docs) that this is sent to the game before it comes here.

src/main/python/rlbot/utils/rendering/rendering_manager.py
```diff
@@ -28,7 +28,7 @@ class RenderingManager:
     bot_index = 0
 
     def __init__(self):
-        self.ignored_funcs = ['setup_function_types', 'get_render_functions',
+        self.ignored_funcs = ['setup_function_types', 'get_rendering_manager',
```
dtracers: We might be able to delete this I do not think this is used anymore

src/main/python/rlbot/botmanager/agent_metadata.py
gyfis: Good point, the stuff about constructors don't really make sense here.

PR #128, opened 2018-06-04
V4 merge candidate

dtracers:


src/main/cpp/RLBotInterface/RLBotInterface/Interface.cpp
```diff
@@ -0,0 +1,164 @@
...
+		if (!FileMappings::Initialize())
+			return ERROR_FUNCTION_FAILED;
+
+		if (!CallbackProcessor::Initialize())
+			return ERROR_FUNCTION_FAILED;
+
+		bInitialized = true;
+		DEBUG_LOG("RLBot Core Interface has been successfully initialized!\n");
+
+		/*while (1)
```
drssoccer55: Clean up commented out test code.

tarehart: This is ccman's, so I'd rather not in this case.

PR #129, opened 2018-06-05
Full struct support and legacy conversion

tarehart:


src/main/cpp/RLBotInterface/RLBotMessages/PacketStructs/MatchDataPacket.hpp
```diff
@@ -3,23 +3,23 @@
...
-	PyStruct::Vector3		LocalMin;
-	PyStruct::Vector3		LocalMax;
-	PyStruct::Vector3		LocalExtent;
-	unsigned char			TeamNum;
+};
+
+struct BoostPad
+{
+	PyStruct::Vector3		Location;
+	bool					FullBoost;
```
dtracers: Spacing

tarehart: Will fix by using spaces instead of tabs for alignment. It lines up in visual studio because it displays tabs as 4 spaces instead of 8.

src/main/python/rlbot/utils/structures/game_data_struct.py
```diff
@@ -121,15 +141,17 @@ def rotate_game_tick_packet_boost_omitted(game_tick_packet):
...
+    game_tick_packet.game_ball.rotation.yaw = ball_yaw + math.pi if ball_yaw < 0 else ball_yaw - math.pi
 
     for i in range(game_tick_packet.num_cars):
         game_tick_packet.game_cars[i].physics.location.x = -1 * game_tick_packet.game_cars[i].physics.location.x
         game_tick_packet.game_cars[i].physics.location.y = -1 * game_tick_packet.game_cars[i].physics.location.y
         game_tick_packet.game_cars[i].physics.velocity.x = -1 * game_tick_packet.game_cars[i].physics.velocity.x
         game_tick_packet.game_cars[i].physics.velocity.y = -1 * game_tick_packet.game_cars[i].physics.velocity.y
-        game_tick_packet.game_cars[i].physics.angular_velocity.x = -1 * game_tick_packet.game_cars[i].physics.angular_velocity.x
-        game_tick_packet.game_cars[i].physics.angular_velocity.y = -1 * game_tick_packet.game_cars[i].physics.angular_velocity.y
+        game_tick_packet.game_cars[i].physics.angular_velocity.x = -1 * game_tick_packet.game_cars[
```
Sciguymjm: Formatting

src/test/python/agents/atba/atba.py
```diff
@@ -52,7 +46,7 @@ def get_output_vector(self, game_tick_packet):
             color = self.renderer.create_color(121, 121, 0, 121)
             color2 = self.renderer.create_color(255, 0, 255, 255)
             self.renderer.draw_line_2d(100, 100, 1000, 500, self.renderer.black())
-            self.renderer.draw_rect_2d(0, 0, 1000, 1000, True, color)
+            self.renderer.draw_rect_2d(0, 0, 50, 50, True, color)
```
Sciguymjm: Are we not getting rid of the default rendering in ATBA? It's a little obnoxious imo. We can disable it by default but leave the code in for reference

tarehart: It's a bit obnoxious but keep in mind that bot makers will never see this since it's in the test directory.

src/main/python/rlbot/utils/structures/game_data_struct.py
tarehart: The auto-formatter did this because the line was too long. I'll break the line at a more appropriate point.

tarehart: Ultimately decided to tell the auto-formatter to shut up instead.

src/test/python/agents/atba/atba.py
Sciguymjm: Ok, that's fine. But the default config still has ATBA as a player for higher than P2. I understand flatBot replaced it, though. Just worried about first time users :^) Otherwise this is fine as a test.

tarehart: Maybe I'll make rendering and quick chatting part of atba's config and turn them both off by default.

PR #132, opened 2018-06-08
Depending on flatbuffers via git instead of storing it in the repo.

tarehart:


tarehart:
I found out setup.py is problematic right now, sorry. Closing the pull request for now, will reopen when it's fixed.

PR #133, opened 2018-06-08
 Applying hallo_doei's GUI fixes

tarehart:


src/main/python/rlbot/gui/preset_editors.py
```diff
@@ -94,7 +94,7 @@ def load_preset_cfg(self):
         file_path = QtWidgets.QFileDialog.getOpenFileName(self, 'Load Config', '', 'Config Files (*.cfg)')[0]
         if not os.path.isfile(file_path):
             return
-        if os.path.splitext(file_path)[1] != "cfg":
+        if os.path.splitext(file_path)[1] != ".cfg":
```
dtracers: we should use https://docs.python.org/3/library/pathlib.html


import pathlib

print(pathlib.Path('yourPathGoesHere').suffix)

Or this and remove the dot

import os.path
extension = os.path.splitext(filename)[1][1:]

src/main/python/setup.py
```diff
@@ -6,8 +6,8 @@
 setuptools.setup(
     name='rlbot',
     packages=setuptools.find_packages(),
-    install_requires=['psutil', 'inputs'],
-    version='0.0.9',
+    install_requires=['psutil', 'inputs', 'PyQt5'],
```
dtracers: why does this not just read from requirements.txt

It is bad to have to change 2 files every time we add a requirement

tarehart: Yeah, good point. I've gotten rid of the duplication by making requirements.txt refer to this.

PR #135, opened 2018-06-09
Making a public_utils module for bot makers to use.

tarehart:
Also improving the sample batch file. RLBotPythonExample will be given something similar.


src/main/python/rlbot/utils/file_util.py
```diff
@@ -24,6 +23,6 @@ def contains_locked_file(directory: str):
                 with open(file_path, 'a'):
                     pass
             except IOError:
-                print(file_path)
+                print("Locked file: {}".format(file_path))
```
dtracers: we should use the logger

PR #141, opened 2018-06-12
fix early invocation of initialize_agent in load_agent function

ericmburgess:
So that the renderer can be accessed as expected from initialize_agent.


tarehart:
Looks good, thanks for doing the pull request! I'm sure it works, but have you had the opportunity to test it yet?

ericmburgess:
I've tested it, works fine.

PR #145, opened 2018-06-21
Adding ability to render centered 3D rectangle. Tidying up rendering.

tarehart:
Part of this change takes place in the core dll.


src/main/python/rlbot/utils/rlbot_exception.py
```diff
@@ -7,10 +7,10 @@ def __init__(self, msg=None):
             msg = "An error occurred attempting to set RLBot configuration on the dll side"
         super(RLBotException, self).__init__(msg)
         self.error_dict = {1: BufferOverfilledError(), 2: MessageLargerThanMaxError(), 3: InvalidNumPlayerError(),
-                           4: InvalidBotSkillError(), 5: InvalidPlayerIndexError(), 6: InvalidName(),
+                           4: InvalidBotSkillError(), 5: InvalidHumanIndex, 6: InvalidName(),
                            7: InvalidTeam, 8: InvalidTeamColor(), 9: InvalidCustomColor, 10: InvalidGameValues,
-                           11: InvalidThrottle, 12: InvalidSteer, 13: InvalidPitch, 14: InvalidRoll,
-                           15: InvalidSteer}
+                           11: InvalidThrottle, 12: InvalidSteer, 13: InvalidPitch, 14: InvalidYaw, 15: InvalidRoll,
+                           16: InvalidPlayerIndexError, 17: InvalidQuickChatPreset, 18: InvalidRenderType}
```
ericmburgess: InvalidPlayerIndexError() is now InvalidPlayerIndexError, (without the parens), is that intended?

tarehart: That's a good point. I refactored a little bit, how do these new changes strike you?

src/main/python/rlbot/utils/structures/game_interface.py
```diff
@@ -121,7 +121,8 @@ def start_match(self):
                                                 None if self.extension is None else self.extension.onMatchStart), None)
 
         if rlbot_status != 0:
-            raise rlbot_exception.RLBotException().raise_exception_from_error_code(rlbot_status)
+            exception_class = get_exception_from_error_code(rlbot_status)
+            raise exception_class()
```
ericmburgess: If you `raise` a class, it'll be instantiated implicitly. So we could do 

```raise get_exception_from_error_code(rlbot_status)```

and it'll have the same effect.

nevercast: That is actually really neat to know.

src/main/python/rlbot/utils/rlbot_exception.py
ericmburgess: Looks great!

PR #148, opened 2018-06-24
Synchronizing java bot registration better.

tarehart:
Tested with four java bots.


tarehart:
Nope, didn't attempt to repro. I ought to have done this in the first place though, I think it's good to have regardless.

nevercast:
Signed off because it's a small fix that is needed regardless. Would be good though to open an issue with the bot dupe glitch and track this commit from there.

nevercast:
I've locked this PR so that all conversation regarding this may take place in the correct place, #149. Any comment you would have left here, please leave there. Thank you!

PR #150, opened 2018-06-25
Rendering support for Java.

tarehart:


src/main/java/rlbot/manager/BotLoopRenderer.java
```diff
@@ -0,0 +1,37 @@
...
+        return botLoopMap.get(bot.getIndex());
+    }
+
+    void startPacket() {
+        builder = new FlatBufferBuilder(1000);
+    }
+
+    void finishAndSendIfDifferent() {
+        RenderPacket packet = doFinishPacket();
+        if (previousPacket == null || !previousPacket.equals(packet)) {
```
nevercast: packet.equals(previousPacket) ties up both of these checks. No?

src/main/java/rlbot/render/NamedRenderer.java
```diff
@@ -0,0 +1,22 @@
...
+public class NamedRenderer extends Renderer {
+    public NamedRenderer(final String renderName) {
+        super(renderName.hashCode());
+    }
+
+    public void startPacket() {
+        builder = new FlatBufferBuilder(1000);
+    }
+
+    public RenderPacket finishPacket() {
```
nevercast: is `finishPacket` going to be used without `finishAndSend()` ? If not, can it be private?

src/main/java/rlbot/render/Renderer.java
```diff
@@ -0,0 +1,225 @@
...
+    private final int index;
+
+    private List<Integer> renderMessageOffsets = new LinkedList<>();
+    protected FlatBufferBuilder builder;
+
+    protected Renderer(int index) {
+        this.index = index;
+    }
+
+    protected RenderPacket doFinishPacket() {
```
nevercast: Any reason not to call this `finishPacket`? Seems inconsistent with `resetPacket`

src/main/java/rlbot/manager/BotLoopRenderer.java
tarehart: Good point!

src/main/java/rlbot/render/NamedRenderer.java
tarehart: Yeah, I wanted to allow the possibility of finishing without sending. People can reuse certain renderings that way.

src/main/java/rlbot/render/Renderer.java
tarehart: Yep, I want the subclass NamedRenderer to be able to have a finishPacket method with public visibility, but I don't want BotLoopRenderer to have any such method exposed to bot makers. My solution was to make this protected, and following from that, I needed to have it be named differently from the finishPacket that exists in NamedRenderer.

nevercast: Awesome! Thanks.

src/main/java/rlbot/render/NamedRenderer.java
nevercast: Nice. Thanks.

nevercast:
Looks like `rlbot.flat.Color` could do with a `createColor(FlatBufferBuilder, java.awt.Color)`. Another time, another PR.

PR #156, opened 2018-07-05
Improve timing accuracy of the main bot loop for C# bots

alion02:


src/main/cs/RLBotDotNet/RLBotDotNet/Manager/BotManager.cs
```diff
@@ -65,18 +66,26 @@ private void RunBot(Bot bot)
...
+                botRunEvent.Set();
+                botRunEvent.Reset();
+
+                // Sleep efficiently (but inaccurately) for as long as we can
+                TimeSpan maxInaccurateSleepTime = targetSleepTime - stopwatch.Elapsed - timerResolution;
+                if (maxInaccurateSleepTime > TimeSpan.Zero)
+                    Thread.Sleep(maxInaccurateSleepTime);
+
+                // Sleep the rest of the time accurately with the use of a tight loop
+                while (stopwatch.Elapsed < targetSleepTime);
```
tarehart: Can you comment on the performance implications here? I'm worried that this will consume a lot of CPU, and might impact people's ability to stream tournaments smoothly, or run other bots on the same team smoothly.

alion02: Assuming a 0.5 ms accurate timer (which is the most common value), this while loop will run for about 3% of sleep time (with the rest handled normally by Thread.Sleep). Omitting this line will make the timings inconsistent, but ticks will never be late (only slightly too early).

tarehart: My personal preference is to omit that line and not risk unreasonable CPU utilization. Have an opinion here, Redox?

TheBlocks: I don't think this empty while loop will perform very efficiently. I think it's better to have slightly early timings rather than having someone's CPU be used up a lot. My opinion here is that the line should be omitted, but if I'm incorrect and it uses up a negligible (or very little) amount of CPU then it's fine to keep.

alion02: Alright, I must agree that it's not critical for this loop to fire at precisely 60hz - it just needs to not be late, for which the while loop is not needed.

Issue #121, opened 2018-06-03
Rendering methods should take in ctypes and FlatBuffers


TheBlocks:
The built-in rendering methods such as `self.renderer.draw_line_3d` should take in the ctypes Vector3 and the FlatBuffers type in addition to tuples/lists for the 3D location arguments.
This would make bot code much cleaner.

TheBlocks:
Merged

PR #166, opened 2018-07-11
Adding rendering support for Flatbuffers and ctypes vectors

TheBlocks:
As per this issue: https://github.com/RLBot/RLBot/issues/121

### DIFF
--- a/src/main/python/rlbot/utils/rendering/rendering_manager.py
+++ b/src/main/python/rlbot/utils/rendering/rendering_manager.py
@@ -214,27 +214,50 @@ class RenderingManager:
 
     """Supports Flatbuffers Vector3, cTypes Vector3, list/tuple of numbers, or passing x,y,z (z optional)"""
     def __create_vector(self, *vec):
-        if (isinstance(vec[0], int) or isinstance(vec[0], float)) and (isinstance(vec[1], int) or isinstance(vec[1], float)):
-            x = vec[0]
-            y = vec[1]
-            if len(vec) == 3 and (isinstance(vec[2], int) or isinstance(vec[2], float)):
-                z = vec[2]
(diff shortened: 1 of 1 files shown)


src/main/python/rlbot/utils/rendering/rendering_manager.py
```diff
@@ -211,7 +212,23 @@ def get_rendering_manager(self, bot_index=0):
     def __wrap_float(self, number):
         return Float.CreateFloat(self.builder, number)
 
-    def __create_vector(self, x, y, z=None):
-        if z is None:
-            z = 0
+    def __create_vector(self, vec):
```
dtracers: it should have support for if the user explictly sent in x,y,z

TheBlocks:
Gotcha. Will add shortly.

src/main/python/rlbot/utils/rendering/rendering_manager.py
```diff
@@ -211,7 +212,29 @@ def get_rendering_manager(self, bot_index=0):
     def __wrap_float(self, number):
         return Float.CreateFloat(self.builder, number)
 
-    def __create_vector(self, x, y, z=None):
-        if z is None:
-            z = 0
+    """Supports Flatbuffers Vector3, cTypes Vector3, list/tuple of numbers, or passing x,y,z (z optional)"""
+    def __create_vector(self, *vec):
+        if (isinstance(vec[0], int) or isinstance(vec[0], float)) and (isinstance(vec[1], int) or isinstance(vec[1], float)):
```
dtracers: you should a length check.

instead of all these instance checks for int and float (also you can check for number)
do a length check for 2 or 3
then later do the check for 3 for z

then you can do the other type checks

PR #172, opened 2018-07-19
add all mutators to the rlbot.cfg file and apply them when starting a match

hallo-doei:
All mutators can now be applied when a match gets started.


src/main/cpp/RLBotInterface/RLBotMessages/PacketStructs/MatchSettings.hpp
```diff
@@ -97,10 +167,68 @@ enum BoostOption
...
+	Two_Seconds,
+	One_Seconds,
+	Disable_Goal_Reset
+};
+
 struct MutatorSettings
 {
 	MatchLength				MatchLength;
+	MaxScore				MaxScore;
+	OvertimeOption			OvertimeOptions;
```
dtracers: these do not line up

src/main/python/rlbot/parsing/match_settings_config_parser.py
```diff
@@ -148,15 +230,35 @@ def get_num_players(config):
...
 def parse_mutator_settings(mutator_settings, config):
     """
     Assigns the mutator settings to the settings object for the dll
     :param mutator_settings:
     :param config:
     """
-    mutator_settings.match_length = match_length_types.index(config.get(MUTATOR_CONFIGURATION_HEADER, MUTATOR_MATCH_LENGTH))
-
-    mutator_settings.boost_options = boost_types.index(config.get(MUTATOR_CONFIGURATION_HEADER, MUTATOR_BOOST_AMOUNT))
+    mutator_settings.match_length = index_def_0(match_length_types, config.get(MUTATOR_CONFIGURATION_HEADER, MUTATOR_MATCH_LENGTH))
```
tarehart: I'm not a fan of silently picking a default when the user enters a bad value. It's almost always better to show them a good error message so they can accomplish what they wanted, and also so they don't get surprised later if we start interpreting their wrong config differently.

PR #173, opened 2018-07-22
Making the match_settings_config_parser backwards compatible

tarehart:
Also fixing a default value for boost strength.


tarehart:
Hold up, this might break the UI, need to do some more testing.

PR #183, opened 2018-08-04
Kill the bots from the gui

dtracers:
does not kill java or c# bots

### DIFF
--- a/src/main/python/rlbot/agents/base_dotnet_agent.py
+++ b/src/main/python/rlbot/agents/base_dotnet_agent.py
@@ -15,6 +15,7 @@ class BaseDotNetAgent(BaseIndependentAgent):
         self.team = team
         self.index = index
         self.port = self.read_port_from_file()
+        self.is_retired = False
 
     def run_independently(self, terminate_request_event):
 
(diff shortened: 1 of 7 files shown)


skyborgff:
i'll make them 
```python
        while True:
            if msvcrt.kbhit():
                return []```
this would be the function that would wait for a keypress right?
you replaced it with a variable changed for the button.
So now the pressing a key will not kill it, that functionality has been deprecated and replaced by the button.

Or will these line keep the "press a key to kill":
```python
        while not self.quit_event.is_set():
            if msvcrt.kbhit():
                msvcrt.getch()
                self.shut_down()```

Is it possible for child.kill() and parent.kill() to fail for a reason other than it being dead already? maybe a successful death should be logged if so.

**Comment by tarehart** [comment]
This code
```
while True:
    if msvcrt.kbhit():
        return []
```
inside the agents was just a hack to avoid those loops preventing a clean shutdown on key press. The primary responsibility for shutting down on key press has always been in setup_manager. As you noticed, this code in setup_manager is still active, we are keeping that functionality.

Using the retire method in the agent is cleaner because there are multiple ways that a shutdown can be initiated (hitting a key, hitting the kill button) and this works for all of them. When the quit event is set, all the agents have retire() called on them.

I'm not sure if child.kill() and parent.kill() can fail for a reason other than it being dead already. However if they do fail, they will emit a big stack trace. Therefore I feel like `self.logger.info("Shut down complete!")` is an adequate way of signaling success.

skyborgff:
Thank you for the explanation :)

PR #186, opened 2018-08-06
Added support for Quick Chats in the C# Library.

whokRL:
The code is based on the Python implementation for Quick Chats.


whokRL:
Preview of the bots using the C# Quick Chat to say "I got it"
quickchat2

Issue #198, opened 2018-08-09
.bat crashes after saving bot.py file in PyCharm


oneunreadmail:
This is what I get every like 5th save if I'm doing it on the go:
`INFO: bot1[      bot_manager.py:154 -                  run() ] Reloading Agent: D:\code\medivh\medivh\medivh.py
Process Process-2:
Traceback (most recent call last):
  File "C:\Users\User\AppData\Local\Programs\Python\Python36\lib\multiprocessing\process.py", line 258, in _bootstrap
    self.run()
  File "C:\Users\User\AppData\Local\Programs\Python\Python36\lib\multiprocessing\process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\User\AppData\Local\Programs\Python\Python36\lib\site-packages\rlbot\setup_manager.py", line 158, in run_agent
    bm.run()
  File "C:\Users\User\AppData\Local\Programs\Python\Python36\lib\site-packages\rlbot\botmanager\bot_manager.py", line 150, in run
    new_module_modification_time = os.stat(agent_class_file).st_mtime
FileNotFoundError: [WinError 2] The system cannot find the file specified: 'D:\\code\\medivh\\medivh\\medivh.py'`

I assume it's because PyCharm makes the file unavailable for reading for a moment, and framework doesn't recheck availability after that.

tarehart:
Good point, that's an annoying problem! Will be fixed soon with https://github.com/RLBot/RLBot/pull/199

PR #199, opened 2018-08-11
Making python hot-reload robust vs. temporary file unavailability.

tarehart:
This helps when using editors like IntelliJ or PyCharm which do some
kind of weird dance when saving files.

This is designed to fix https://github.com/RLBot/RLBot/issues/198

### DIFF
--- a/src/main/python/rlbot/botmanager/bot_manager.py
+++ b/src/main/python/rlbot/botmanager/bot_manager.py
@@ -147,19 +147,23 @@ class BotManager:
                 last_call_real_time = datetime.now()
 
                 # Reload the Agent if it has been modified.
-                new_module_modification_time = os.stat(agent_class_file).st_mtime
-                if new_module_modification_time != last_module_modification_time:
-                    try:
-                        last_module_modification_time = new_module_modification_time
+                try:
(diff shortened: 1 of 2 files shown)


src/main/python/rlbot/botmanager/bot_manager.py
```diff
@@ -147,19 +147,23 @@ def run(self):
                 last_call_real_time = datetime.now()
 
                 # Reload the Agent if it has been modified.
-                new_module_modification_time = os.stat(agent_class_file).st_mtime
-                if new_module_modification_time != last_module_modification_time:
-                    try:
-                        last_module_modification_time = new_module_modification_time
+                try:
+                    new_module_modification_time = os.stat(agent_class_file).st_mtime
```
dtracers: could do an if check for existence instead of wrapping this all in a try catch?
I know that is not pythonic but it feels cleaner

tarehart: I'd rather do it this way partly because we don't have to worry about a race condition where the existence check check passes, then the file gets locked or moved, and the os.stat fails.

PR #201, opened 2018-08-12
All Mutators in the GUI

hallo-doei:
The GUI now has a popup to edit all the mutators, it gets saved together with the overall_config, autosave is included if it is checked in the main window.


twobackfromtheend:
image

PR #218, opened 2018-08-19
Updating to patch-proof SDK version. Improving ATBA rendering.

tarehart:


TheBlocks:
With this update, it seems that shutting down RLBot in the command line and then starting it again causes RLBot to start a new match instead of resuming the ongoing match.
Is this intentional? I can see this being annoying when the user accidentally shuts down RLBot but they don't want to start a new match.

tarehart:
Good catch! I have a fix for the core dll that's currently under review. When it's approved, I'll update this PR with a recompiled dll.

PR #230, opened 2018-08-23
Ability to force reload of all bots by pressing 'r'. Press 'q' to quit.

NicEastvillage:
Works by passing an event to BotManager which will make it reload when set.
BotManager still reloads on file modification.
Made sure it worked, when a human is playing.
You have to press 'q' to quit now.
More commands should be easy to add in the future.

### DIFF
--- a/src/main/python/rlbot/setup_manager.py
+++ b/src/main/python/rlbot/setup_manager.py
@@ -45,7 +45,6 @@ class SetupManager:
         self.helper_process_manager = HelperProcessManager(self.quit_event)
         self.bot_quit_callbacks = []
         self.bot_reload_requests = []
-        self.bot_reload_callbacks = []
         self.agent_metadata_map = {}
 
     def startup(self):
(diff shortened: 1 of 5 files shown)


src/main/python/rlbot/setup_manager.py
```diff
@@ -111,12 +117,21 @@ def run(self):
...
         while not self.quit_event.is_set():
+            # Handle commands
             if msvcrt.kbhit():
-                msvcrt.getch()
-                self.shut_down()
-                break
+                command = msvcrt.getwch()
+                if command.lower() == 'r':  # reload
+                    self.reload_all_agents()
+                elif command.lower() == 'q':  # quit
```
tarehart: Could you listen for ESC also? Doesn't need to be in the instructions, I think it'll be convenient to have though.

NicEastvillage:
You're right, I didn't know what to do on reload fail, so now it just does nothing. I will delete the reload_complete event and add ESC as shut down key.

PR #246, opened 2018-09-04
Setting gamestate in C#

kipje13:
It works but could be more user friendly.


src/main/cs/RLBotDotNet/RLBotDotNet/GameState/GameState.cs
```diff
@@ -0,0 +1,68 @@
...
+            DesiredGameState.AddCarStates(builder, carStateVector);
+
+            builder.Finish(DesiredGameState.EndDesiredGameState(builder).Value);
+
+            return new GameStatePacket(builder.SizedByteArray());
+        }
+
+        private static VectorOffset CreateCarStateVector(FlatBufferBuilder builder, Dictionary<int, CarState> carStates)
+        {
+            int numCars = carStates.Keys.Max() + 1;
```
tarehart: Does this work if carStates is empty?

src/main/cs/RLBotDotNet/RLBotDotNet/GameState/DesiredVector3.cs
```diff
@@ -0,0 +1,42 @@
+﻿using rlbot.flat;
+using FlatBuffers;
+
+namespace RLBotDotNet.GameState
+{
+    public class DesiredVector3
```
tarehart: Please add a constructor that takes zero arguments. Currently the only way to make the ball drift at Z = 300 is to construct with x, y, and z, then go back and set x and y to null.

Also, would you mind adding some test code to TestBot?

tarehart:
That'll do, I'll add some code to TestBot in a separate PR. Thanks kipje!

PR #251, opened 2018-09-07
Adding the ability to take a snapshot of the game in C# + more user friendly state setting.

kipje13:
I also added some code to the example bot based on @tarehart 's PR but with the updated getters and constructors.


tarehart:
Since this pull request lists a bunch of commits from old PRs, can you please look at the "files changed" tab and confirm that these are the exact changes you intend to make?

src/main/cs/RLBotDotNet/RLBotDotNet/Interface/RLBotInterface.cs
```diff
@@ -164,7 +173,17 @@ public static void SetGameStatePacket(GameStatePacket gameState)
...
             byte[] bytes = finishedRender.Bytes;
-            RenderGroup(bytes, bytes.Length);
+            int status = RenderGroup(bytes, bytes.Length);
+
+            if (status > 0)
+                throw NewRLBotCoreException((RLBotCoreStatus)status);
+        }
+
+
+        private static RLBotCoreException NewRLBotCoreException(RLBotCoreStatus status)
```
tarehart: What's your reasoning for making this static helper method instead of another constructor for RLBotCoreException?

kipje13: As far as I'm aware you can only set the Message field of an exception by passing a string to the constructer. So if I would have made a constructor taking a RLBotCoreStatus and then generate a string, there would be no way of setting that Message field.

tarehart: We discussed offline about ways to do this. Since I can't reach you at the moment, I'm going to go ahead and merge this then follow up with another commit.

PR #260, opened 2018-09-25
New dll with a fix for dodges.

tarehart:


tarehart:
Folks with permissions can see the changed code here: https://github.com/tarehart/rlbot-core/pull/26

PR #270, opened 2018-10-06
Feature/physics tick

whatisaphone:
This grabs values directly from the physics data instead of the tick data

Obligatory demo video: https://discordapp.com/channels/348658686962696195/352507627928027138/497938399332663307

In the video, the game is slowed down to 0.1 by BakkesMod, and then the video is slowed down by another 1/4 so you can clearly see the difference, for a total slowdown of 40x. The top number is the game tick time, and the rest are physics data changing at the equivalent of 120Hz when sped up (so 120/40 = 3 physics ticks per second in the video).

Python flatbuffers is slow enough that sometimes it misses ticks at full speed (with game set to 250fps), but that shouldn't be a problem in any other language (besides Excel 😄).

You can reproduce the test video (I hope) by building/copying DLLs and flatbuffers, setting test_physics_tick = True in atba.cfg, and running runner.bat.

Companion PR: https://github.com/tarehart/rlbot-core/pull/31


src/main/python/rlbot/agents/base_agent.py
```diff
@@ -96,6 +98,10 @@ def get_field_info(self):
         This does not change during a match so it only needs to be called once after the everything is loaded."""
         return self.__field_info_func()
 
+    def get_physics_tick(self) -> PhysicsTick:
```
dtracers: Considering we just added a native version for ball prediction we might need that for physics as well.

If the point of this is instead for ML recording (which this would be great for) it might be good to do the "saving" of the data in the interface dll.

src/main/flatbuffers/rlbot.fbs
```diff
@@ -158,6 +165,31 @@ table GameTickPacket {
...
+/// The state of a rigid body in RL's physics engine.
+table RBState {
+  frame:int;
+  quaternion:Quaternion;
+  location:Vector3;
+  velocity:Vector3;
+  angularVelocity:Vector3;
+}
+
+table PlayerPhysicsInfo {
```
dtracers: I feel like making this "RawPlayerPhysicsInfo" to make it stand out.

src/main/flatbuffers/rlbot.fbs
```diff
@@ -158,6 +165,31 @@ table GameTickPacket {
...
+  state:RBState;
+  input:ControllerState;
+}
+
+table BallPhysicsInfo {
```
dtracers: Same here with Raw so it is different then the other 2 versions of this

tarehart:
Oh man, this might win PR of the year!

whatisaphone:
> Oh man, this might win PR of the year!

I've been here less than a week and I've already peaked 😅

src/main/flatbuffers/rlbot.fbs
tarehart: I'm conflicted about the name, but some `///` documentation definitely wouldn't go amiss (which gets translated to javadoc, etc when classes are generated).

tarehart:
Looks great! I'll add a commit for struct conversion once the .fbs naming is agreed upon.

src/main/flatbuffers/rlbot.fbs
whatisaphone: Can do comments.

I think this is more deserving of the name "Physics" than the others, but we can't go back and change the old names of course.

I don't care about the shed color, someone decree a list of names and I'll change them to whatever

src/main/python/rlbot/agents/base_agent.py
whatisaphone: I assume by "native" you mean C structs – Tare's on it.

Saving the data can happen wherever anyone likes, that can be a separate PR.

src/main/flatbuffers/rlbot.fbs
```diff
@@ -158,6 +165,31 @@ table GameTickPacket {
 
 root_type GameTickPacket;
 
+/// The state of a rigid body in RL's physics engine.
+table RBState {
+  frame:int;
+  quaternion:Quaternion;
```
tarehart: I'm thinking about changing this to rotation:Quaternion. Any objections?

whatisaphone: If you're going to do that, can you move it below location as well? e.g. the field order will become:

```
table RBState {
  frame:int;
  location:Vector3;
  rotation:Quaternion;
  velocity:Vector3;
  angularVelocity:Vector3;
}
```

(Because we shouldn't mix conventions)

tarehart: Wilco.

src/main/cpp/RLBotInterface/RLBotMessages/PacketStructs/RigidBodyStructs.hpp
```diff
@@ -0,0 +1,43 @@
...
+	float W;
+};
+
+struct RigidBodyState
+{
+	PyStruct::Vector3 		Location;
+	Quaternion 				Rotation;
+	PyStruct::Vector3		Velocity;
+	PyStruct::Vector3		AngularVelocity;
+	unsigned long			Frame;
```
whatisaphone: I'd have put Frame first so it matches the other definition, but it's not a huge deal

src/main/flatbuffers/rlbot.fbs
```diff
@@ -41,13 +41,16 @@ struct Vector3 {
...
-// Values are in radians
+/// Expresses the rotation state of an object in Euler angles, with values in radians.
 struct Rotator {
   pitch:float;
   yaw:float;
   roll:float;
 }
 
+/// Expresses the rotation state of an object.
+/// x, y, and z can be considered an axis of rotation, and w the amount of rotation.
```
whatisaphone: This comment isn't quite accurate. Example with w=0:

src/main/python/rlbot/utils/structures/game_interface.py
```diff
@@ -277,16 +277,11 @@ def get_live_data_flat_binary(self):
             self.game_status(None, RLBotCoreStatus.Success)
             return proto_string
 
-    def get_physics_tick(self) -> PhysicsTick:
+    def update_rigid_body_tick(self, rigid_body_tick: RigidBodyTick):
```
whatisaphone: Do you think it's worth keeping both implementations here for testing purposes? (maybe suffix the FlatBuffer one with `_flatbuffer`)

src/main/cpp/RLBotInterface/RLBotMessages/PacketStructs/RigidBodyStructs.hpp
whatisaphone: Also, to match the RL definition, it should just be `int`.

src/main/python/rlbot/utils/structures/rigid_body_struct.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class Quaternion(ctypes.Structure):
+    _fields_ = [("x", ctypes.c_float),
+                ("y", ctypes.c_float),
+                ("z", ctypes.c_float),
+                ("w", ctypes.c_float)]
+
+
+class RigidBodyState(ctypes.Structure):
```
whatisaphone: tagging this since there are comments about it above

src/main/cpp/RLBotInterface/RLBotMessages/PacketStructs/RigidBodyStructs.hpp
tarehart: Good point, will fix both! I guess I'll watch and see what happens when it rolls over...

src/main/python/rlbot/utils/structures/game_interface.py
tarehart: I'd rather not, unless we identify a production use case. Everything we add to python goes into a gray area where it's a quasi-public API. It's not necessarily documented as a public API, but there's nothing stopping people from finding it and using it, so it becomes uncomfortable to change. I'd like to keep that surface area small if possible.

PR #281, opened 2018-10-26
Failing fast with specific errors when cfg files are messed up.

tarehart:
There are lots of different cfg problems that manifest as a "Could not locate a suitable bot class" error, when really the user just has an invalid file path. This should give accurate error messages.

Also better in general to fail fast rather than confuse people with bot behavior or loadout configs that they don't expect, e.g. atba looks and not moving.


tarehart:
Thanks for telling me to check for GUI crashes, hallo. It turns out that when running it in debugger mode in IntelliJ it never crashes, but when running it normally it did!

PR #282, opened 2018-11-05
Core dll with fixed fieldInfo for dropshot.

tarehart:
Fixed a bug where field info was not extracted properly during dropshot mode. It was reporting 2 goals rather than the expected 140.

Change in core dll: https://github.com/skyborgff/rlbot-core/pull/3


hallo-doei:
Good job finding the cause! Let's get some great dropshot bots

PR #291, opened 2018-11-18
Launch game on clicking run if running process not found (QoL)

ima9rd:
I've added some code to launch Rocket League when the run button is clicked if the process is not found using psutil. This includes a variable defining Rocket League's Steam gameid and process name(s). Also added a utility in process_configuration file to check if a process is running by name.

### DIFF
--- a/src/main/python/rlbot/gui/qt_root.py
+++ b/src/main/python/rlbot/gui/qt_root.py
@@ -7,6 +7,7 @@ from PyQt5.QtGui import QIcon
 from PyQt5.QtWidgets import QFileDialog, QMainWindow, QApplication, QWidget, QComboBox, QLineEdit, QRadioButton, QSlider, QCheckBox, QMessageBox
 import threading
 import configparser
+import webbrowser
 
 from rlbot.gui.presets import AgentPreset, LoadoutPreset
 from rlbot.gui.index_manager import IndexManager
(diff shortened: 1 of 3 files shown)


ima9rd:
Wasn't certain on the best place to put the utility and whether or not to make the process information a variable or just include it directly in the code, so let me know if anyone has any preferences on this.

src/main/python/rlbot/gui/qt_root.py
```diff
@@ -98,6 +102,8 @@ def run_button_pressed(self):
             # Attempting to run again when we're in this state can result in duplicate processes.
             return
         self.launch_in_progress = True
+        if not is_process_running(ROCKET_LEAGUE_PROCESS_INFO['program'], ROCKET_LEAGUE_PROCESS_INFO['program_name']):
```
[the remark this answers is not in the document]
ima9rd: Moved the code to setup_manager.py as per Skyborg's recommendation.

skyborgff:
Doesn't this add another dependency or is web a built-in package?

ima9rd:
> Doesn't this add another dependency or is web a built-in package?

I believe this is a built in module - https://github.com/python/cpython/tree/master/Lib/webbrowser.py

src/main/python/rlbot/utils/process_configuration.py
```diff
@@ -71,3 +71,18 @@ def extract_all_pids(agent_metadata_map):
     for player_index, data in agent_metadata_map.items():
         pids.update(data.pids)
     return pids
+
+
+def is_process_running(program, scriptname):
+    if not optional_packages_installed:
+        return True
+    for pid in psutil.pids():
```
slysz: Consider using `psutil.process_iter` instead, it would be less memory heavy and a bit faster as you would not need to call `psutil.Process` on every pid

aseeon: Consider using `psutil.process_iter` instead, it would be less memory heavy and a bit faster as you would not need to call `psutil.Process` on every pid

PR #294, opened 2018-11-21
Add autostart to C#/.NET bots.

kipje13:


tarehart:
Nice, looks similar to https://github.com/RLBot/RLBot/commit/d0494f02c240f37a01c7ca4959d9fd5452b1343a in all the right ways. I might have left some useless imports in the Java one, whoops!

tarehart:
I tested the python aspects of this and it's working well!

I don't have Visual Studio installed at the moment to test with an exe, but I assume you've already done that right? If so, please go ahead and add version notes, then I'll approve / merge / deploy. This can be 1.6.0.

kipje13:
Yes, I have tested auto starting the .exe, aswell as having a wrong path or no path at all.

Issue #287, opened 2018-11-16
BaseSubprocessAgent relative path issue


whatisaphone:
I realized a deficiency in `BaseSubprocessAgent` while helping people set up for the tourney yesterday. Right now, it looks for the configured exe implicitly relative to the process cwd. Instead, it should look for it explicitly relative to the path of the `.cfg` file that contains the path. That way, you can move the directory containing the .cfg and .exe wherever you want, and it will still work. This should be fixed before the next tournament.

tarehart:
You're in luck because I just added a "getpath" function to the config header class which can be used right here: https://github.com/RLBot/RLBot/blob/master/src/main/python/rlbot/agents/base_subprocess_agent.py

It will resolve paths relative to the config file. https://github.com/RLBot/RLBot/blob/master/src/main/python/rlbot/parsing/custom_config.py

whatisaphone:
That is very relevant, looks like it will be a 4-character diff then!

ima9rd:
Should be fixed now 👍

whatisaphone:
Thanks!

PR #295, opened 2018-11-22
Fix GUI preset load and relative path problems

ima9rd:
This fixes GUI loading to not discard preset configuration files that share the same name but are located in different directories. The GUI would correctly save the rlbot.cfg file when multiple agents were added with the same name, but the loading code would overwrite all presets loaded after the initial due to matching on preset name alone.

### DIFF
--- a/src/main/python/rlbot/agents/base_subprocess_agent.py
+++ b/src/main/python/rlbot/agents/base_subprocess_agent.py
@@ -18,11 +18,11 @@ class BaseSubprocessAgent(BaseIndependentAgent):
         params.add_value('path', str, description='The path to the exe for the bot')
 
     def load_config(self, config_header: ConfigHeader) -> None:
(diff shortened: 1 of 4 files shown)


ima9rd:
I'm guessing there exists code somewhere to correctly append the suffix for presets, but I couldn't find it (likely related to debugging for 3 hours straight to find the source of this issue)

ima9rd:
Added in another GUI crash fix for agents outside of the default directory relative path.

ima9rd:
More commits to fix relative path issues.

ima9rd:
Unfortunately the base_subprocess_agent didn't fix the issue whatisaphone was experiencing. I'm going to continue to work on this.

whatisaphone:
I can confirm this fixes #287.

ima9rd:
Seems OK to merge now based on testing 👍

PR #296, opened 2018-11-22
Make RLBot work on Unix

martmists:


tarehart:
Thanks for chipping in! Unfortunately getting RLBot to work on Unix also involves a lot of work in other parts of the code, especially the part where we inject logic into the game.

That part is not open source, and we don't have plans to attempt it in the short term since it will be extremely difficult (not sure if it's even possible), and we don't get very many Unix users.

martmists:
Ah, understandable. If you do want to support Unix users, I'd be happy to help out. I don't have windows installed at all, so implementing this would be interesting to say the least.

cominixo01:
Yeah, I'm in the same situation, only linux installed and I don't want to dual boot just for RLBot

tarehart:
Closing for now since there's a lot of deeper changes needed first for Unix.

PR #297, opened 2018-11-23
Fixed the issue where participant_team would be a boolean value instead of int in the config

jeroen11dijk:
I had the issue where participant_team would sometimes be set to a boolean value after closing the gui. I asked on the discord and someone set I should do the casting on the two functions I changed. After which I have not encountered the error anymore but since I dont know anything about the framework Im not entirely sure that i have fixed it.

Also I needed two commits to fix the config back to normal

### DIFF
--- a/rlbot.cfg
+++ b/rlbot.cfg
@@ -1,3 +1,26 @@
+Skip to content
+ 
+Search or jump to…
+
+Pull requests
+Issues
+Marketplace
+Explore
+ @jeroen11dijk Sign out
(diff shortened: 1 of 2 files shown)


tarehart:
I believe I fixed this problem with https://github.com/RLBot/RLBot/pull/286/files

Have you seen any evidence that this change is needed also?

jeroen11dijk:
No I don't so I think you have fixed it with that request. But I havent used the GUI as intens as I did when I tested all the hopshot entries where I had to swap a lot. But for now I will close this one.

PR #298, opened 2018-11-27
Updated to include python 3.7 as official

dtracers:


README.md
```diff
@@ -22,7 +22,7 @@ It provides values from the game like car and ball position, and carries back bu
 RLBot works for up to 10 bots.
 
 ### Requirements
-Windows, Rocket League, Python 3.6.
+Windows, Rocket League, Python 3.7, 3.6 (deprecated)
```
TheBlocks: I think it would be better writing `Python 3.7 (recommended) or 3.6` instead of saying that 3.6 is deprecated, because it's not.

nevercast: Whatever is decided should be replicated here https://github.com/RLBot/RLBotPythonExample/pull/15

skyborgff: So, will this be done? This is specially important now that the new GUI requires 3.7

dtracers: Are we dropping 3.6?

skyborgff: From my understanding we should, as the new gui uses a fixed version

tarehart: The GUI comes bundled with python 3.7 and doesn't care what people have pre-installed, so it's still no problem if people use 3.6 for the old GUI or when developing bots.

I'd still prefer not to declare 3.6 as deprecated, since it works fine and is not causing any maintenance or support burden.

TheBlocks: Agreed with tarehart, but I'd still like to encourage people to use 3.7 over 3.6. Instead of saying that 3.6 is deprecated, we should instead say that 3.7 is recommended.

skyborgff:
one like and i merge

README.md
tarehart: I think that's a good idea.

martmists:
LGTM

skyborgff:
OH GOD YES

PR #302, opened 2018-12-03
December fix

skyborgff:


tarehart:
We have a tiny little git situation here. You can fix it by fetc

tarehart:
I'm an idiot

PR #303, opened 2018-12-06
Don't scare people by logging to info about locked files.

tarehart:
People have assumed that this is the root cause of their error about 9000 times. Using logger.debug makes it not appear in the console anymore (currently we only log at info level).


nevercast:
It is times like this I'm glad we don't have a `node_modules` in the local project. Imagine walking that tree! 😨

PR #304, opened 2018-12-06
Add a Python version check in case the user is running an old version

DomNomNom:
Hello, guess who's back?

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -1,6 +1,7 @@
 import sys
 import os.path
 sys.path.insert(0, os.path.realpath(os.path.dirname(os.path.abspath(__file__)) + '/src/main/python/'))
+from rlbot.utils import python_version_check
 from rlbot import runner as framework_runner
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


IamEld3st:
Personally... I think there should still be a warning for using 3.6 as we plan to switch to 3.7

tarehart:
Welcome back Dom!

I like this change as-is. I'd personally rather not issue warnings about 3.6 since it will continue to work fine for the foreseeable future. 

The only incentives for for "standardizing" on a single version that I can think of:
- Allow chip's RLUtilities library to only target one, but:
  - Not everybody uses it
  - RLUtilities users will find out about the additional constraint when they adopt it
  - RLUtilities will hopefully achieve some low-effort way of targeting multiple versions
- Reduce the scope of our testing / troubleshooting for the main rlbot, but:
  - That's lame
  - Supporting both 3.6 and 3.7 is causing zero trouble for the main rlbot so far
- Reduce the likelihood that somebody will develop a bot under different conditions than it will be run under during a tournament, but:
  - It's the tournament runner's prerogative to choose and announce what python version they'll be using, and the entrants will probably actually read that.

By the way I had no idea you could compare tuples like that and it's blowing my mind :)

nevercast:
I just did some version checking yesterday in python because `argparse` in 3.6 is borked, and I did not know of this tuple comparison either. Nutty!

PR #307, opened 2018-12-09
Support numpy arrays for rendering.

RLMarvin:


IamEld3st:
From what I see vec3 is chips vector and ndarray is numpy. I would keep them separated.

tarehart:
Yep we're seeing the same thing. I don't see a reason to duplicate the code though, they will both support index access forever. I don't like duck typing usually, but it's handy here.

whatisaphone:
Peanut gallery chiming in. I'd keep them separated as well, it's cleaner that way. Think about something like Kotlin type narrowing, where it type checks each block with different type information. Or V8, which would JIT each block according to the type check. Those technical reasons don't apply in Python, but they all build off the general principle of having a well-typed program on every line, and that does still apply imo.

whatisaphone:
The thing that jumps out to me more is using string comparison for the type checks. I think those should be `isinstance` calls.

RLMarvin:
I think to use isinstance you'll have to import the type, so you'd need to have those installed.
I could just merge the two if statements into one line, or make a try except block if we care about supporting any subscriptable type that contains numbers.

PR #309, opened 2018-12-19
Make logging more configurable.

RLMarvin:
At the moment we can't add a handler with a level lower than the default logging_level, for example if i want to only show info logs in console but write debug level logs to a file.

We should have the level of the logger itself to be as low as possible and only set a default for the console for example.

This allows:
 - adding a handler with any level.
 - configuring the default handler (self.logger.handlers[0]) without changing the root logger and affecting all loggers.

### DIFF
--- a/src/main/python/rlbot/agents/base_agent.py
+++ b/src/main/python/rlbot/agents/base_agent.py
@@ -60,7 +60,7 @@ class BaseAgent:
         self.name = name
         self.team = team
         self.index = index
-        self.logger = get_logger('nameless_bot')
+        self.logger = get_logger(f'bot{index}')
 
     def get_output(self, game_tick_packet: GameTickPacket) -> SimpleControllerState:
         """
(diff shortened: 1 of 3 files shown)


nevercast:
"Allow custom handlers for logging, enabling logging debug information to file"
"Separate configuration of default bot logger so that it doesn't trample settings of other loggers"

nevercast:
Would be good to have a really basic example file for this, @RLMarvin are you up for showing debug logging to file? I know anyone that reads Python logging docs can figure it out, but it would be good to have.

RLMarvin:
> 
> 
> Would be good to have a really basic example file for this, @RLMarvin are you up for showing debug logging to file? I know anyone that reads Python logging docs can figure it out, but it would be good to have.



It's as simple as adding the following:
```
file_handler = logging.FileHandler(file_name)
self.logger.addHandler(file_handler)
```
I don't think this has to go in the framework though.

PR #310, opened 2018-12-19
Automatically clamping player inputs between -1 and 1, and issuing warnings

tarehart:
Before this change, an invalid value for an input, e.g. a throttle of 1.5, could cause a car to stop moving. That's bad. The framework would also not print any error messages in that situation. That's worse.

After this change, a throttle of 1.5 will be treated as 1.0 and the car will drive just fine. Also, even though the car will be driving fine, we'll log warnings so the bot maker can double check their logic.

I tested this by setting the ATBA throttle to 1.2 and watching it drive around while spamming warnings in the console.

Although this change touches the core and interface dlls, this PR will **NOT** include fresh binaries or version notes. That's because we have another change coming through very shortly which will also affect the binaries, and it's more space-efficient in the repo to only update the binaries once.


src/main/cpp/RLBotInterface/RLBotInterface/GameFunctions/PlayerInfo.cpp
```diff
@@ -92,21 +92,34 @@ namespace GameFunctions
...
 		if (status != RLBotCoreStatus::Success)
 			return status;
 
-		flatbuffers::FlatBufferBuilder builder;
-		FlatbufferTranslator::inputStructToFlatbuffer(&builder, playerInput, playerIndex);
-		return UpdatePlayerInputFlatbuffer(builder.GetBufferPointer(), builder.GetSize());
+		// We're validating the input *after* sending it to the core dll because the core dll
+		// is going to clamp it for us with no issues. We're reporting validation errors to
+		// users just for their information because it might help them discover faulty logic.
+		return checkInputConfiguration(playerInput);
```
ccman32: Seems a little weird to me to validate it afterwards but I guess it makes sense for reporting errors to the user

src/main/cpp/RLBotInterface/RLBotInterface/GameFunctions/PlayerInfo.cpp
```diff
@@ -92,21 +92,34 @@ namespace GameFunctions
...
 	}
 
 	// FLAT
 	static BoostUtilities::QueueSender flatInputQueue(BoostConstants::PlayerInputFlatQueueName);
 
 	extern "C" RLBotCoreStatus RLBOT_CORE_API UpdatePlayerInputFlatbuffer(void* controllerState, int protoSize)
 	{
-		return flatInputQueue.sendMessage(controllerState, protoSize);
+		RLBotCoreStatus status = flatInputQueue.sendMessage(controllerState, protoSize);
```
ccman32: Does sendMessage suddenly return a RLBotCoreStatus?

PR #100, opened 2018-05-31
Flat quick chat

dtracers:
Quick chat is now done through flat buffers.

Callbacks are still not implemented correctly

PR #312, opened 2018-12-20
Fix custom quick chats.

RLMarvin:


src/main/cpp/RLBotInterface/RLBotMessages/MessageStructs/GameMessages.hpp
```diff
@@ -55,6 +55,21 @@ enum QuickChatPreset
...
+	Custom_Toxic_DeAlloc,
+	Custom_Toxic_404NoSkill,
+	Custom_Toxic_CatchVirus,
+	Custom_Useful_Passing,
+	Custom_Useful_Faking,
+	Custom_Useful_Demoing,
+	Custom_Useful_Bumping,
+	Custom_Compliments_TinyChances,
+	Custom_Compliments_SkillLevel,
+	Custom_Compliments_proud,
```
whatisaphone: If this is the first time these names will be exposed outside the core, maybe we should capitalize on our chance to capitalize this correctly
```suggestion
 	Custom_Compliments_Proud,
```

nevercast: I am in agreement.

nevercast:
Let's get @whatisaphone's change in there and then go ahead with the merge. @dtracers Have you tested that quick chat is now working?

src/main/cpp/RLBotInterface/RLBotMessages/MessageStructs/GameMessages.hpp
RLMarvin: [https://github.com/RLBot/RLBot/blob/master/src/main/flatbuffers/rlbot.fbs#L386](url)

nevercast: That line gives me a `git blame` for here: https://github.com/RLBot/RLBot/pull/100
@dtracers Can you confirm what breaks if we fix both files?

tarehart: This is not the first time the capitalization is exposed outside of core. If anybody is currently using the proud quickchat, they'll have `Custom_Compliments_proud` in their code and this would be a breaking change for them, manifesting as a compilation error in some languages.

The impact would probably be one or two people at most, but Marvin is under no obligation to take crap for making a breaking change. If it were me, I'd just leave it alone.

PR #313, opened 2018-12-20
Sorting the customisation items in the gui

hallo-doei:
This commit only changes the order in which the items get added to the combo boxes. The order in the box being changed doesn't need any other adjustments since it checks for the text and not for the index when selecting an item.


skyborgff:
unknown is for manually input of ids that we dont have a "name" for?

hallo-doei:
Unknown has always been there and is indeed the text the combobox will show when you input an id which is not in the list

skyborgff:
Maybe not in this commit, but we should add a detection for known ids that belong to other cathegories

skyborgff:
!redox, i'll just accept :P

nevercast:
Since this doesn't have a version bump in it, I'll add to the 1.7.1 milestone so we can keep up with change logs when it lands in Pip

PR #316, opened 2018-12-23
Pass a few more options to the child process

whatisaphone:
The big one is `--rlbot-dll-directory`. This will save tourney organizers from having to futz with `PATH`


whatisaphone:
I'm pretty sure Rust is the only consumer. There's been murmuring about bots in C++ and Node which would use this, but I haven't seen anyone actually do it yet.

(Also my argument parser is far from sane, but don't tell anyone 😄)

nevercast:
Chip and I have discussed starting works on a C++ framework in the New Year, but until then, this is Rust only.

PR #318, opened 2018-12-28
Fix double logs.

RLMarvin:
pipmain() (used in run.py on the example bots) seems to add a handler to the root logger which resutls in a doubling of logs :

```
INFO:rlbot[   game_interface.py:196 -           inject_dll() ] Injecting DLL
Injecting DLL
```

This is a fix to that.


tarehart:
Thanks for the fix!

PR #319, opened 2018-12-28
Add common python functionality to game data structs.

DomNomNom:
Example use:

    >>> from rlbot.utils.structures.game_data_struct import Touch
    >>> Touch() == Touch()
    True
    >>> Touch()
    Touch<player_name: '', time_seconds: 0.0, hit_location: Vector3<x: 0.0, y: 0.0, z: 0.0>, hit_normal: Vector3<x: 0.0, y: 0.0, z: 0.0>>

This is intended to help newcomers explore the datastructures they're reading.
Also to allow people to more easily compare whether two structs are the same (useful for testing).

### DIFF
--- a/src/main/python/rlbot/utils/structures/game_data_struct.py
+++ b/src/main/python/rlbot/utils/structures/game_data_struct.py
@@ -2,42 +2,42 @@ import ctypes
 import math
 
 from rlbot.utils.structures.start_match_structures import MAX_NAME_LENGTH, MAX_PLAYERS
-
 from rlbot.utils.structures.utils import create_enum_object
+from rlbot.utils.structures.struct import Struct
 
 MAX_BOOSTS = 50
 MAX_TILES = 200
 MAX_GOALS = 200
 
 
-class Vector3(ctypes.Structure):
+class Vector3(Struct):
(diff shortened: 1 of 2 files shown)


nevercast:
Will python implicitly use `__eq__` if `__ne__` is undefined? Otherwise, can `__ne__` just return `not __eq__(self,  other)`, duplicate implementation may be a course of errors. It also seems that if a struct is similar to another they will be equal. i.e.

Point(x, y) == Rect(x, y, w, h); however, Rect(x, y, w, h) != Point(x, y). Equality here is not commutative. It should be. 

This is because you are only iterating the fields of one struct, not a zip of the two.

DomNomNom:
@nevercast Yes, python does seem to fall back to `not __eq__()` when `__ne__()` isn't defined. 
 I added type checking to make sure equality is commutative. Iterating over a `zip` of attributes would lead to non-commutativity again.

Current behaviour:

    >>> from rlbot.utils.structures.game_data_struct import Touch, GameInfo
    >>> Touch() == GameInfo()
    False
    >>> Touch() == Touch()
    True

nevercast:
This is perfect, thanks for the swift fix on that @DomNomNom

PR #326, opened 2018-12-31
Faster initialization by ccman

tarehart:
The injected dll will now initialize *very* quickly. Thanks ccman!

This corresponds to https://github.com/skyborgff/rlbot-core/pull/38


src/main/cpp/RLBotInterface/RLBotMessages/MutexUtilities/MutexUtilities.cpp
```diff
@@ -0,0 +1,94 @@
+#include "MutexUtilities.hpp"
+
+#include <Windows.h>
```
tarehart: One day when we support Linux I'll replace this with a boost mutex >:)

This looks good though!

PR #327, opened 2018-12-31
Add GameSpeed to GameInfo and state setting.

RLMarvin:


src/main/flatbuffers/rlbot.fbs
```diff
@@ -157,6 +157,7 @@ table GameInfo {
   /// countdown. Turns false again the moment the 'choose team' screen appears.
   isMatchEnded:bool;
   worldGravityZ:float;
+  gameSpeed:float;
```
dtracers: can you add a comment explaining what unit of time gamespeed is?

is it frames, seconds, milliseconds?

PR #333, opened 2019-01-02
Switch to using TimeDilation for game speed.

RLMarvin:
This change was made since using the previous method we couldn't speed up the game only slow it down, now that's fixed!

Here is a clip showcasing the effects of a speedup: https://gfycat.com/UltimateAcidicBlackwidowspider


tarehart:
Very nice!

Can you please go ahead and bump the version number in version.py? Also can you confirm that this core dll includes ccman's compatibility fix from here? <https://github.com/skyborgff/rlbot-core/commit/eb2ff39b78dd86656ca8e7c6d04eb5db33728b09>

I foolishly deleted the rlbot-core branch before checking.

RLMarvin:
> 
> 
> Very nice!
> 
> Can you please go ahead and bump the version number in version.py? Also can you confirm that this core dll includes ccman's compatibility fix from here? [skyborgff/rlbot-core@eb2ff39](https://github.com/skyborgff/rlbot-core/commit/eb2ff39b78dd86656ca8e7c6d04eb5db33728b09)
> 
> I foolishly deleted the rlbot-core branch before checking.

Yup it does include it.

PR #337, opened 2019-01-05
Fixing a bug with indexed values in config objects.

tarehart:
I noticed this while working on RLBotGUI

### DIFF
--- a/src/main/python/rlbot/parsing/custom_config.py
+++ b/src/main/python/rlbot/parsing/custom_config.py
@@ -155,15 +155,15 @@ class ConfigHeader:
         """
         Sets the value on the given option.
         :param option: The name of the option as it appears in the config file
-        :param value: The value that is being applied, if this section is indexed value must be a list
+        :param value: The value that is being applied. If this section is indexed then the
+        value must be a list (to be applied directly) or you must supply the index parameter,
+        which will cause the value to be inserted into an existing list.
+        :param index: If the attribute is indexed, we will use this index to insert
+        the value you have supplied.
         :return: an instance of itself so that you can chain setting values together.
         """
-        # Should raise error if indexed and there's no list or if indexed and no index given
-        if self.is_indexed and index is None:
-            if not isinstance(value, list):
-                raise TypeError("Value should be a list when not giving an index in an indexed header")
-            else:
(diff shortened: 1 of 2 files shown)


src/main/python/rlbot/version.py
```diff
@@ -4,18 +4,19 @@
...
-    '1.9.3': """
+    '1.9.4': """
     - *Much* faster core dll initialization! - ccman32
     - Adding support for a training mode! Check out https://github.com/RLBot/RLBotTraining - DomNomNom
     - Allow the user to change the appearance of human and party-member bot agents via the GUI - r0bbi3
     - Added game speed info to game tick packet and the ability to modify it via state setting - Marvin
     - Make the game stop capturing the mouse cursor if only bots are playing - whatisaphone
     - Various quality-of-life improvements - DomNomNom
     - Making the GUI load all bots in the current directory - tarehart
+    - Fixing a bug with indexed values in config objects - tarehart
```
DomNomNom: Should we make a new section instead?

src/main/python/rlbot/parsing/custom_config.py
```diff
@@ -155,15 +155,15 @@ def set_value(self, option, value, index=None):
         """
         Sets the value on the given option.
         :param option: The name of the option as it appears in the config file
-        :param value: The value that is being applied, if this section is indexed value must be a list
+        :param value: The value that is being applied. If this section is indexed then the
+        value must be a list (to be applied directly) or you must supply the index parameter,
+        which will cause the value to be inserted into an existing list.
+        :param index: If the attribute is indexed, we will use this index to insert
```
DomNomNom: No changes required in this PR, I just want to note that I find it a bit confusing to understand what the `is_indexed` attribute stands for upon first reading this file.

src/main/python/rlbot/version.py
tarehart: I like to make sure big contributions are on display for at least a week.

DomNomNom: Acknowledged.

src/main/python/rlbot/parsing/custom_config.py
tarehart: Good point, I just pushed a small improvement to the documentation.

PR #339, opened 2019-01-05
Match configuration refactoring to make new GUIs easier.

tarehart:
Facilitates stuff like this:

image


skyborgff:
With this do you mean custom guis?
could you be a little more precise please?

tarehart:
The PR is about how we deal with match and player settings in the back end.

It makes rlbot support a much easier way of specifying the players and match settings without dealing with the config file format.

PR #340, opened 2019-01-06
Change default appearance.

RLMarvin:


src/main/python/rlbot/agents/base_agent.py
```diff
@@ -270,33 +270,33 @@ def base_create_agent_configurations(cls) -> ConfigObject:
...
+        header.add_value('trails_id', int, default=3220, description='Car trail Selection')
+        header.add_value('goal_explosion_id', int, default=3018, description='Goal Explosion Selection')
 
         return header
 
     @staticmethod
     def _create_loadout_paint() -> ConfigHeader:
         header = ConfigHeader()
 
-        header.add_value('car_paint_id', int, default=0)
```
tarehart: I suspect that changing this default will have an effect on any bot which does not specify paint config, which would be bad. I haven't tested though.

RLMarvin: You're right it does, my bad.

PR #341, opened 2019-01-06
Using MatchConfig as the main match starting interface, and passing it to bots

tarehart:
This will allow bots to know what map they're playing on, what mutators, etc.

This also makes our python code more elegant imo.

### DIFF
--- a/src/main/python/rlbot/matchconfig/match_config.py
+++ b/src/main/python/rlbot/matchconfig/match_config.py
@@ -1,7 +1,6 @@
 from typing import List
 
 from rlbot.matchconfig.loadout_config import LoadoutConfig
-from rlbot.parsing.bot_config_bundle import BotConfigBundle
 from rlbot.parsing.match_settings_config_parser import boost_amount_mutator_types, map_types, game_mode_types, \
     match_length_types, max_score_types, overtime_mutator_types, series_length_mutator_types, game_speed_mutator_types, \
     ball_max_speed_mutator_types, ball_type_mutator_types, ball_weight_mutator_types, ball_size_mutator_types, \
@@ -25,7 +24,7 @@ class PlayerConfig:
         self.human_index: int = None
         self.name: str = None
         self.team: int = None
-        self.config_bundle: BotConfigBundle = None  # Required only if rlbot_controlled is true
+        self.config_path: str = None  # Required only if rlbot_controlled is true
         self.loadout_config: LoadoutConfig = None
 
     def write(self, player_configuration: PlayerConfiguration, name_dict: dict):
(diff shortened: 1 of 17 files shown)


src/main/python/rlbot/matchconfig/match_config.py
```diff
@@ -104,7 +118,27 @@ def create_match_settings(self) -> MatchSettings:
...
+    :param name: The name that is being sanitized
+    :return: A sanitized version of the name
+    """
+    if name not in dict:
+        new_name = name[:31]  # Make sure name does not exceed 31 characters
+        dict[name] = 1
+    else:
+        count = dict[name]
+        new_name = name[:27] + "(" + str(count + 1) + ")"  # Truncate at 27 because we can have up to '(10)' appended
+        dict[name] = count + 1
```
DomNomNom: Have you considered collections.Counter?

src/main/python/rlbot/matchconfig/match_config.py
```diff
@@ -66,22 +66,33 @@ def __init__(self):
...
+        mutator_settings.ball_bounciness_option = self.index_or_zero(ball_bounciness_mutator_types, self.ball_bounciness)
+        mutator_settings.boost_amount_option = self.index_or_zero(boost_amount_mutator_types, self.boost_amount)
+        mutator_settings.rumble_option = self.index_or_zero(rumble_mutator_types, self.rumble)
+        mutator_settings.boost_strength_option = self.index_or_zero(boost_strength_mutator_types, self.boost_strength)
+        mutator_settings.gravity_option = self.index_or_zero(gravity_mutator_types, self.gravity)
+        mutator_settings.demolish_option = self.index_or_zero(demolish_mutator_types, self.demolish)
+        mutator_settings.respawn_time_option = self.index_or_zero(respawn_time_mutator_types, self.respawn_time)
+
+    @staticmethod
+    def index_or_zero(types, value):
```
DomNomNom: "types" seems like a tricky thing to understand when first reading this.
How about refactoring our lists to enums such that we have a better type signature. e.g.
`def index_or_zero(enum_type: Type[Enum], value: Enum) -> int:`

Changing things to enums is fine even if we have names with spaces. e.g. 
```
Station = Enum(
    value='Station',
    names=[
        ('wien_westbahnhof', 1),
        ('Wien Westbahnhof', 1),
        ('st_poelten', 2),
        ('St. Pölten', 2),
        ('linz', 3),
        ('Linz', 3),
    ]
)
```

If we are changing things to enums, we could make things even more convenient by putting the index_or_zero on a class our Enums inherit from such that the above lines look like this:
`mutator_settings.demolish_option = DemolishMutator.parse_or_default(self.demolish)`
I'd find this approach much more readable.

Issue #324, opened 2018-12-30
Add global team scores to game tick packet


NicEastvillage:
There is no reason not to add them to the game tick packet. It's just a few bytes and they can be read directly from the game like all the other values.

nevercast:
Duplicate: https://github.com/RLBot/RLBot/issues/323

Issue #323, opened 2018-12-30
game_interface is missing information - blocking implementation of training


DomNomNom:
It is currently impossible to always find out when there was a goal. Adding this information to `GameInfo.score_info` would be sufficient. I've tried to work around it but code along the lines of `sum(car.score_info.goals for car in game_tick_packet)` does not produce the correct answers as there is no attribution to players if there is a goal without a player touching the ball in Goalie training.

PR #344, opened 2019-01-09
Add Team Score.

RLMarvin:
This adds number of goals scored by each team.
Potential fix for #323 & #324


tarehart:
Awesome, thanks Marvin! I'll approve this as soon as I get home to test it.

whatisaphone:
Awesome, we've been needing this for a while!

What do you think about using an array instead of two separate fields?

RLMarvin:
I think it could be more future-proof, as in if we add team specific info it'll have structure already in place.
However it adds unnecessary complexity if it's an array that'll only ever hold one value.

What's your thoughts?

tarehart:
You both have good points. I feel like there are two separate decisions here:
- Do we want to start a struct for team specific info
  - Pleasantly extensible, as noted above
  - Mildly more complex
- Do we want to express the data as an array or two specific fields
  - An array would allow people to index into it more elegantly with their bot's team number. Some people would probably also consider an array to be better aesthetically.

RLMarvin:
Thanks for the clarification!
I'm fine with doing both, but that may delay the feature a bit.

tarehart:
If you think the answer to both is 'yes', I'm happy to pitch in.

whatisaphone:
I skimmed through the `ATeam*` fields and I don't think there's anything else in there besides score that can change in the middle of a game. But there's lots of mildly interesting data that should be constant during a game. The ones that jump out at me are TeamName/TeamColor (useful for UIs or tournament runners?), ForfeitVote (let bots forfeit!), TemporarySpawnSpots and/or SecondaryStarts (spawn positions after a demo). If you can see a future where we add any of those, let's do a struct, otherwise not.

I like your idea of moving it up to `GameTickPacket`.

RLMarvin:
Thanks for the feedback! I'll try making the changes, hopefully this goes out before ep6!

src/main/cpp/RLBotInterface/RLBotMessages/MessageTranslation/FlatbufferTranslator.cpp
```diff
@@ -156,6 +164,14 @@ namespace FlatbufferTranslator {
 			}
 		}
 
+		auto teams = flatPacket->teams();
+		if (teams) {
+			packet->NumTeams = teams->size();
+			for (int i = 0; i < teams->size() && i < CONST_MaxTeams; i++) {
+			fillTeamStruct(teams->Get(i), &packet->Teams[i]);
```
DomNomNom: indentation

src/main/cpp/RLBotInterface/RLBotMessages/MessageTranslation/FlatbufferTranslator.cpp
```diff
@@ -156,6 +164,14 @@ namespace FlatbufferTranslator {
 			}
 		}
 
+		auto teams = flatPacket->teams();
+		if (teams) {
+			packet->NumTeams = teams->size();
+			for (int i = 0; i < teams->size() && i < CONST_MaxTeams; i++) {
```
DomNomNom:
```suggestion
			packet->NumTeams = min(teams->size(), CONST_MaxTeams);
			for (int i = 0; i < packet->NumTeams; i++) {
```

src/main/python/rlbot/utils/structures/game_data_struct.py
```diff
@@ -145,34 +154,52 @@ class ByteBuffer(Struct):
 # This negates all x and y values for balls and cars and rotates yaw 180 degrees.
 def rotate_game_tick_packet_boost_omitted(game_tick_packet):
     # Negate all x,y values for ball
-    game_tick_packet.game_ball.physics.location.x = -1 * game_tick_packet.game_ball.physics.location.x
-    game_tick_packet.game_ball.physics.location.y = -1 * game_tick_packet.game_ball.physics.location.y
-    game_tick_packet.game_ball.physics.velocity.x = -1 * game_tick_packet.game_ball.physics.velocity.x
-    game_tick_packet.game_ball.physics.velocity.y = -1 * game_tick_packet.game_ball.physics.velocity.y
+    game_tick_packet.game_ball.physics.location.x = - \
+        1 * game_tick_packet.game_ball.physics.location.x
```
DomNomNom: The  `-` being on a different line to the `1` makes it easy to mis-read by missing the negation.
Put the `-` on the new line too.

RLMarvin: This was the fault of my auto-formatted :(

PR #356, opened 2019-01-15
Fix the halving of the update frequency for bots and add the option to set a frequency

kipje13:


src/main/cs/RLBotDotNet/RLBotDotNet/Manager/BotManager.cs
```diff
@@ -19,16 +19,27 @@ namespace RLBotDotNet
...
         private Thread serverThread;
+
+        private readonly int frequency;
         
         /// <summary>
         /// Constructs a new instance of BotManager.
         /// </summary>
         public BotManager()
         {
             _renderers = new ConcurrentDictionary<int, BotLoopRenderer>();
```
tarehart: Is this a good opportunity for constructor chaining? https://stackoverflow.com/questions/1814953/c-sharp-constructor-chaining-how-to-do-it

src/main/cs/RLBotDotNet/RLBotDotNet/Manager/BotManager.cs
```diff
@@ -115,15 +129,18 @@ private void MainBotLoop()
...
+                {
+                    proc.botRunEvent.Set();
+                }
 
                 // Sleep efficiently (but inaccurately) for as long as we can
                 TimeSpan maxInaccurateSleepTime = targetSleepTime - stopwatch.Elapsed - timerResolution;
                 if (maxInaccurateSleepTime > TimeSpan.Zero)
                     Thread.Sleep(maxInaccurateSleepTime);
 
                 // We could sleep the rest of the time accurately with the use of a spin-wait, but since the main bot loop doesn't have to fire at precise intervals it's reasonable to omit this step
```
tarehart: This comment seems out of date now.

Does the spin wait have an adverse effect on overall CPU utilization, e.g. could it hurt the performance of Rocket League, twitch streaming, or other bots on the team? (in theory bots on the other team are shielded by CPU affinity)

src/main/cs/RLBotDotNet/RLBotDotNet/Manager/BotManager.cs
kipje13: Thanks for the link! Didn't know about that. Changed it and also added a bounds check on the frequency to stop errors from occuring later down the line.

src/main/cs/RLBotDotNet/RLBotDotNet/Manager/BotManager.cs
kipje13: I found about 1% CPU usage in the diagnostics tool. I could add another parameter to the constructor that allows people to enable the spin wait. This will also help with backwards compatibility.

PR #357, opened 2019-01-17
Adding the new Details config section to base agent configurations.

tarehart:
The config values are based on what @NicEastvillage added to the standard configs here: https://github.com/RLBot/RLBotPythonExample/pull/14

This will let me read the values from bundle.base_agent_config without getting an error about 'Details' not being a header. The error was happening even for bots who defined it in their config.

Also getting rid of an error message that tends to scare people.

### DIFF
--- a/src/main/python/rlbot/parsing/rlbot_config_parser.py
+++ b/src/main/python/rlbot/parsing/rlbot_config_parser.py
@@ -70,11 +70,12 @@ def parse_configurations(config_parser: ConfigObject, config_location, config_bu
         match_config.player_configs.append(player_config)
 
     extension_path = config_parser.get(RLBOT_CONFIGURATION_HEADER, EXTENSION_PATH_KEY)
-    if extension_path:
+    if extension_path and extension_path != 'None':  # The string 'None' ends up in people's config a lot.
         match_config.extension_config = ExtensionConfig()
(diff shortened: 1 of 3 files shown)


src/main/python/rlbot/parsing/rlbot_config_parser.py
```diff
@@ -70,12 +70,13 @@ def parse_configurations(config_parser: ConfigObject, config_location, config_bu
         match_config.player_configs.append(player_config)
 
     extension_path = config_parser.get(RLBOT_CONFIGURATION_HEADER, EXTENSION_PATH_KEY)
-    if extension_path:
+    if extension_path and extension_path != 'None':  # The string 'None' ends up in people's config a lot.
```
DomNomNom: Could we also fix the source of this problem please?

tarehart:
Yep, multi-line config was tested when it was originally added to the RLBot*Example repos. Thanks for reviewing!

src/main/python/rlbot/parsing/rlbot_config_parser.py
tarehart: Yes I'd like to fix that too. Gonna save it for a future PR though since we're gonna need this little band-aid regardless.

Issue #358, opened 2019-01-17
Incorrect timing between get_output calls in python


kipje13:
When running a python bot with the fps set to 120, the time between two get_output calls is approximatly 10ms as can be seen here:

![](https://cdn.discordapp.com/attachments/348659150793736193/535428760888147968/unknown.png)

At 120fps we should be expecting a time of about 8ms. Since we are considerably over that limit, we are dropping about 1 in 5/6 frames.

When running a python bot at 60 fps, the time between get_output calls is approximatly 20ms. We should be expecting 16ms in this case.

![](https://cdn.discordapp.com/attachments/348659150793736193/535431891487162378/unknown.png)

As can be seen here, we run 2/3 frames at 20ms, followed by a 10ms frame. This Presumably happens to catch up from the time lost on the previous frames. 

This will limit the effective time budget to 10ms. When going over 10ms, you will drop about 1 in 3/4 frames.

ddthj:
This is stale, and has a much cooler variant over here: #445

PR #359, opened 2019-01-19
Added rendering capability to training exercises

DomNomNom:
Also added testing for the training API.
I have refactored the existing integration test, but it is still failing as intended. (https://github.com/RLBot/RLBot/issues/358)


src/main/python/rlbot/training/status_rendering.py
```diff
@@ -1,10 +1,22 @@
 from collections import namedtuple
+from contextlib import contextmanager
 from typing import List
 
 from rlbot.utils.rendering.rendering_manager import RenderingManager
 
 Row = namedtuple('Row', 'exercise_name status_str status_color_func')
 
+@contextmanager
+def training_status_renerer_context(exercise_names: List[str], renderman: RenderingManager):
```
RLMarvin: Little typo here, and the other places where it's used.

DomNomNom: Thanks for spotting that. Fixed.

tarehart:
I tested a decent amount and everything is working, including the integration test. Great stuff!

PR #363, opened 2019-01-22
Clear the screen when bots retire

DomNomNom:
"No single corner human can occupy or experience more than a single corner at the same time during a 4-corner rotation within the 4/16 creation principle." 
  \- The wisest human, Dr. Gene Ray


ddthj:
Yay

src/main/python/rlbot/version.py
```diff
@@ -15,6 +15,7 @@
     - Making the details section of bot config files more visible to python GUIs. - tarehart
     - Improved stability of the SetupManager. - DomNomNom
     - Added rendering capability to training exercises. - DomNomNom
+    - Clear the screen when bots retire. - DomNomNom
```
tarehart: The version number also needs to be bumped, otherwise pypi will reject us. I'll go ahead and do that.

PR #367, opened 2019-01-26
Fix for serialization of multiline values in config files

tarehart:
Parsing this then writing it again resulted in the second line not being indented: https://github.com/RLBot/RLBotPythonExample/blob/master/python_example/python_example.cfg#L17-L18

Subsequent parsing of the config file fails.

Also catching the ParsingError which occurs from this in the directory scanner.


tarehart:
Once this is deployed, I can let this feature off the chain https://www.youtube.com/watch?v=Vx3Bfvq5Ro4

PR #370, opened 2019-01-29
Flatbuffer spec for starting matches.

whatisaphone:
This is @tarehart's branch and it's still in the oven, I'm just making it a PR so it's easier to keep track of


whatisaphone:
🎉 

Is this ready for review/testing, or still WIP?

tarehart:
Still WIP, there's another translation commit I haven't sent yet, plus I want to write a test in Java. Should be soon!

src/main/flatbuffers/rlbot.fbs
```diff
@@ -292,171 +292,418 @@ table DesiredGameState {
...
+  antennaPaintId: int;
+  hatPaintId: int;
+  trailsPaintId: int;
+  goalExplosionPaintId: int;
+}
+
+table PlayerConfiguration {
+  playerClass: PlayerClass;
+  name: string;
+  team: ubyte;
```
DomNomNom: This is inconsistent with TinyPlayer which stores this as an int.
How about storing it as an enum in both cases?

src/test/java/rlbot/cppinterop/RLBotDllTest.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class RLBotDllTest {
+
+    @BeforeAll
+    static void initialize() throws IOException {
+        RLBotDll.initialize("src/main/python/rlbot/dll/RLBot_Core_Interface.dll");
+
```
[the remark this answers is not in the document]
tarehart: Not intended, I'll get rid of it. The java communities I've been a part of don't have rules for newlines, so I just do what I think is readable. That one looked bad, so fair play.

src/main/flatbuffers/rlbot.fbs
tarehart: Good call. It's also stored as an int in PlayerInfo, so I'll be consistent with that. I'm not keen on switching to an enum in those existing tables, because we'd have to leave the int around as a deprecated field to avoid breaking backwards compatibility. Plus all the bots use int for team and they're not going to migrate any time soon.

PR #371, opened 2019-02-02
 Add MIT license that shows up in `pip show rlbot` 

DomNomNom:


src/main/python/rlbot/version.py
```diff
@@ -4,9 +4,13 @@
 # 3) we can import it into your module module
 # https://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package
 
-__version__ = '1.12.8'
+__version__ = '1.13.2'
 
 release_notes = {
+    '1.13.2': """
+    - Fix accidental publishing of pypi package 1.13.1 - DomNomNom
```
tarehart: Technically only 1.13.0 was accidental, 1.13.1 was me un-breaking things :)

How about `Clumsily starting version 1.13 by mistake. - tarehart`?

PR #373, opened 2019-02-11
Making java wait for interface dll initialization

tarehart:
Related pull request for core: https://github.com/skyborgff/rlbot-core/pull/46

Now is_match_ended will be a pretty reliable indicator. It becomes true after the replay of the final goal, right before the celebration phase.

Also includes an improvement to dll initialization for java bots that might fix a few issues.

### DIFF
--- a/src/main/java/rlbot/cppinterop/RLBotDll.java
+++ b/src/main/java/rlbot/cppinterop/RLBotDll.java
@@ -39,8 +39,9 @@ public class RLBotDll {
     private static native ByteBufferStruct GetBallPrediction();
     private static native int SendQuickChat(Pointer ptr, int size);
     private static native int StartMatchFlatbuffer(Pointer ptr, int size);
-    private static native boolean IsInitialized();
+    private static native boolean IsInitialized();  // This asks the dll instance whether it is done initializing.
 
+    // This helps us keep track internally of whether we have initialized, so that we only do it once.
     private static boolean isInitializationComplete = false;
     private static final Object fileLock = new Object();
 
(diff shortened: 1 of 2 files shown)


src/main/java/rlbot/cppinterop/RLBotDll.java
```diff
@@ -77,6 +78,14 @@ public static void initialize(final String interfaceDllPath) throws IOException
             System.out.println("Loading DLL from " + dllSource);
             Native.register(dllNameSansExtension);
 
+            while (!IsInitialized()) {
```
DomNomNom: Maybe add a maximum amount of time that we wait for?

src/main/java/rlbot/cppinterop/RLBotDll.java
```diff
@@ -39,6 +39,7 @@
     private static native ByteBufferStruct GetBallPrediction();
     private static native int SendQuickChat(Pointer ptr, int size);
     private static native int StartMatchFlatbuffer(Pointer ptr, int size);
+    private static native boolean IsInitialized();
 
     private static boolean isInitialized = false;
```
DomNomNom: Is this still needed? Feels like it could lead to typo errors.
If you do decide to keep both, please document the difference.

src/main/java/rlbot/cppinterop/RLBotDll.java
tarehart: Good idea, done.

src/main/java/rlbot/cppinterop/RLBotDll.java
tarehart: Yep, I still want both. Renamed and documented.

Issue #386, opened 2019-02-28
Subprocess bots kill the GUI after a game


whatisaphone:
[11:48 AM] tarehart: Hey, I was checking out SDC, and I noticed that it broke RLBotGUI upon retiring
[11:48 AM] tarehart: The GUI perceived a keyboard interrupt and also exited

…

[12:19 PM] tarehart: presumably [base_subprocess_agent](https://github.com/RLBot/RLBot/blob/17c487d069c3c997fcd92962a4a1a50c7d2d1600/src/main/python/rlbot/agents/base_subprocess_agent.py#L38) is the culprit:

```python
terminate_request_event.wait()
process.send_signal(CTRL_C_EVENT)
process.wait()
```

[12:20 PM] whatisaphone: that should be going to the child process, not the current process
[12:20 PM] whatisaphone: although maybe it propagates and I didn't realize
[12:21 PM] whatisaphone: it shouldn't though. if you open a command prompt and hit ctrl-c, it doesn't violently kill windows explorer
[12:22 PM] tarehart: I find the apparent propagation surprising too

cc @tarehart

PR #387, opened 2019-02-28
Sending SIGTERM instead of CTRL_C_EVENT when subprocess agents retire.

tarehart:
This is designed to fix https://github.com/RLBot/RLBot/issues/386

Self-driving car was killing the GUI when it retired. SIGTERM appears not to have that issue.


tarehart:
Oh, and I tested that SIGTERM actually kills the process, too. When using the kill bots button, SDC stopped playing as expected.

tarehart:
Ah, interesting. I'd like to improve on the situation, but for now I'm gonna take your approval and run with it :)

PR #390, opened 2019-03-10
Don't wait on Psyonix bots to say they're ready to start an exercise.

DomNomNom:
Because only RLBot bots post metadata


DomNomNom:
This should also make it possible to use human players in rlbottraining.

PR #391, opened 2019-03-15
Document that we support up to 64 players now

DomNomNom:


skyborgff:
Could you add the info to the GUI as well?

DomNomNom:
> Could you add the info to the GUI as well?

Done.

PR #394, opened 2019-03-24
Wait for bot respawn in training

DomNomNom:


src/main/python/rlbot/training/training.py
```diff
@@ -177,16 +177,18 @@ def _wait_until_good_ticks(game_interface: GameInterface, required_new_ticks: in
...
-        # Read from game data shared memory
-        game_interface.update_live_data_packet(game_tick_packet)
-        tick_game_time = game_tick_packet.game_info.seconds_elapsed
-        if tick_game_time != last_tick_game_time and game_tick_packet.game_info.is_round_active:
-            last_tick_game_time = tick_game_time
+        game_interface.update_live_data_packet(packet)
+        def is_good_tick():
+            if packet.game_info.seconds_elapsed == last_tick_game_time: return False
+            if not packet.game_info.is_round_active: return False
+            if any(car.is_demolished for car in packet.game_cars): return False
```
tarehart: I was briefly worried that this would hinder an AdversityBot test where you demo several enemies in some allotted time, but I see I was mistaken. Now I just want to write the test :)

DomNomNom: The scenario you described should be fine. This is only waiting before an exercise starts. Once the exercise begins, demos can happen and can be detected by the exercises' Grader.

The reason for this wait is that state setting fails when the bots are in a demolished state. There is a minor concern when before an exercise starts, Adversity is demoing a bot in the frame where we see all opponents being alive but when we state-set then the bots are not alive.
That case I can't work around other than asking for better state setting code.

PR #396, opened 2019-04-06
KILL. TERMINATE. EXTERMINATE. End bot processes when training is done.

DomNomNom:
https://youtu.be/e59guruVL4o

### DIFF
--- a/src/main/python/rlbot/setup_manager.py
+++ b/src/main/python/rlbot/setup_manager.py
@@ -49,7 +49,7 @@ def setup_manager_context():
     try:
         yield setup_manager
     finally:
-        setup_manager.shut_down()
+        setup_manager.shut_down(kill_all_pids=True)
 
 class SetupManager:
     """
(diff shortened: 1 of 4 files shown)


src/main/python/rlbot/setup_manager.py
```diff
@@ -49,7 +49,7 @@ def setup_manager_context():
     try:
         yield setup_manager
     finally:
-        setup_manager.shut_down()
+        setup_manager.shut_down(kill_all_pids=True)
```
dtracers: did you test this to make sure that in normal use this does not cause any issues?

DomNomNom: I've tested it with RLBotGUI as well as RLBotTraining.
The normal (GUI) way of running RLBot does not use this contextmanager at the moment.

tarehart:
Plus I'm strangely relaxed now

PR #418, opened 2019-04-26
Minor changes for portability

PixelyIon:
This commit adds in some minor changes to allow for portability. However, some complex use cases such as those in FileMappings.cpp and Interface.cpp may require entirely different segments of code for Windows and Linux as interoperable functions are missing in many cases.

### DIFF
--- a/src/main/cpp/RLBotInterface/RLBotMessages/FileMappings/FileMappings.cpp
+++ b/src/main/cpp/RLBotInterface/RLBotMessages/FileMappings/FileMappings.cpp
@@ -51,8 +51,7 @@ namespace FileMappings
 
 	void Lock(void* pSharedMem)
 	{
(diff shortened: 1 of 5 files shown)


src/main/cpp/RLBotInterface/RLBotMessages/FileMappings/FileMappings.cpp
```diff
@@ -51,8 +51,7 @@ namespace FileMappings
 
 	void Lock(void* pSharedMem)
 	{
-		while (InterlockedCompareExchange((unsigned int*)pSharedMem, LOCKED, UNLOCKED) != UNLOCKED)
-			Sleep(0);
+		while (InterlockedCompareExchange((unsigned int*)pSharedMem, LOCKED, UNLOCKED) != UNLOCKED) {}
```
tarehart: The sleep was probably significant: https://stackoverflow.com/questions/3727420/significance-of-sleep0

Moot point though, this file is gone in 1.15+ :)

PR #446, opened 2019-08-25
Botmanager py36 fix

ericmburgess:
This should fix the backward incompatibility with Python 3.6 that my previous PR introduced.


ericmburgess:
In my testing, `perf_counter` and `perf_counter_ns` have the same precision--they just use different units. I like the _ns version because its name is kind of self-documenting, but when I made that choice I hadn't realized it was new in 3.7.

If we anticipate dropping 3.6 support soon, then I like this way because it's a simple change to remove the shim. But if that's not happening soon (or maybe ever?) then it's probably cleaner just to rewrite. What do you think?

tarehart:
That's fair enough I suppose, thanks for the fix!

PR #452, opened 2019-09-27
Allowing python bots to configure their preferred maximum tick rate.

tarehart:
I want to give python bots the ability to specify their maximum_tick_rate_preference.

If they set maximum_tick_rate_preference = 60 then get_output will get called at most 60 times per second.

I intend to:
- Set this value to 60 by default
- Inform our power users who want 120 that they need to update their .cfg file

This is a sensible default because the majority of our bot makers have been developing and testing on 60 Hz. I'm confident of this because few people even know how to find and modify TARLBot.ini to achieve 120.

I've tested this code pretty well. When the setting is 120, I've never observed the bot manager skipping a frame. The in-game percentages overlay responds as expected. If you set maximum_tick_rate_preference = 12, it shows a steady 10% when RLBot.exe is updating data at 120.

I'll update version.py properly in the closed source repo; this branch is a bit out of date.

### DIFF
--- a/src/main/python/rlbot/agents/base_agent.py
+++ b/src/main/python/rlbot/agents/base_agent.py
@@ -22,6 +22,7 @@ LOGO_FILE_KEY = 'logo_file'
 LOOKS_CONFIG_KEY = 'looks_config'
 BOT_NAME_KEY = "name"
 SUPPORTS_EARLY_START_KEY = "supports_early_start"
+MAXIMUM_TICK_RATE_PREFERENCE_KEY = "maximum_tick_rate_preference"
 
 
 class SimpleControllerState:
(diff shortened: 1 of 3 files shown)


skyborgff:
This probably will need a overhaul of %, or a better definition for it.
should we always consider 100% as 120hz? 
or should we consider 100% whatever the bot expects as 100%?
i would prefer the first option, but currently, if we play at 60fps, 100% is 60hz

tarehart:
A definition that currently works: Out of the number of frames made available by RLBot.exe, the bot has given a response for X% of them.

I like the way it works now
- Going over 100% is always cause for concern, regardless of language or config setting
- People are already used to Java bots showing 50%

skyborgff:
maybe add a hz count, like:
60hz
100%
98%

or
120hz
96%
50%

so we can know what "physicsrate" the user is getting

PR #453, opened 2019-09-29
Allow Java bots to specify their Refresh Rate

GodGamer029:
Also fixes a minor error in DesiredVector3.java

### DIFF
--- a/src/main/java/rlbot/manager/BotManager.java
+++ b/src/main/java/rlbot/manager/BotManager.java
@@ -88,29 +88,58 @@ public class BotManager {
     }
 
     private void doLoop() {
-        long timeOfLastRefresh = System.currentTimeMillis();
+        // Minimum call rate when paused.
+        final long MAX_AGENT_CALL_PERIOD = 1000 / 30;
+
+        long rateLimitTime = System.currentTimeMillis();
+
+        float lastTickGameTime = -1;
+        float frameUrgency = 0;
+        long lastCallRealTime = System.currentTimeMillis();
         while (keepRunning) {
-            try {
-                timeOfLastRefresh = System.currentTimeMillis();
+            // Python version: https://github.com/RLBot/RLBot/blob/master/src/main/python/rlbot/botmanager/bot_manager.py#L194-L212
+            final int refreshRate = this.refreshRate.get();
+            try{
+                // Retrieve latest packet
                 latestPacket = RLBotDll.getFlatbufferPacket();
-                synchronized (dinnerBell) {
-                    dinnerBell.notifyAll();
+
+                // Run the bot only if gameInfo has updated
+                final float tickGameTime = latestPacket.gameInfo().secondsElapsed();
+                final long now = System.currentTimeMillis();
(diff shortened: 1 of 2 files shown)


src/main/java/rlbot/manager/BotManager.java
```diff
@@ -98,8 +100,7 @@ private void doLoop() {
             }
 
             try {
-                // Fetch the latest game tick packet at 60 Hz.
-                Thread.sleep(16);
+                Thread.sleep(1000 / refreshRate.get());
```
skyborgff: If we sleep exactly the time between frames it will eventually skip a frame because of the compute time outside the sleep state.

GodGamer029: Temporary fix in e66c9ea

GodGamer029: Might cause duplicate frames now though

GodGamer029:
Full fix might be to calculate the current packet/second rate and choose the `Thread.sleep` based on that, but i personally think that is a bit overkill

src/main/java/rlbot/manager/BotManager.java
```diff
@@ -86,22 +88,60 @@ public void ensureStarted() {
     }
 
     private void doLoop() {
-        while (keepRunning) {
+        // Minimum call rate when paused.
+        final long MAX_AGENT_CALL_PERIOD = 1000 / 30;
+        final float tarehartsConstant = 2f;
```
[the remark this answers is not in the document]
GodGamer029: Having the value at 1 kind of defeats the point of having it in the first place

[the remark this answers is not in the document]
skyborgff: angry_tarehart_ava.png

src/main/java/rlbot/manager/BotManager.java
```diff
@@ -86,22 +88,48 @@ public void ensureStarted() {
...
                 latestPacket = RLBotDll.getFlatbufferPacket();
-                synchronized (dinnerBell) {
-                    dinnerBell.notifyAll();
+
+                // Run the bot only if gameInfo has updated
+                final float tickGameTime = latestPacket.gameInfo().secondsElapsed();
+                final long now = System.currentTimeMillis();
+                final boolean shouldCallWhilePaused = now - lastCallRealTime >= MAX_AGENT_CALL_PERIOD;
+
+                if(frameUrgency < TAREHARTS_CONSTANT / refreshRate) // Urgency increases every frame, but don't let it build up a large backlog
```
tarehart: These lines still seem redundant to me. I believe you could replace these two lines with just `frameUrgency = Math.min(frameUrgency + (tickGameTime - lastTickGameTime), TAREHARTS_CONSTANT / refreshRate);`

I might be wrong though. I'm about to come through and add https://github.com/RLBot/RLBot/commit/3d4ccfeb5f306679ea574e9daa76c0d4521783a0, so I'll experiment with it at that point.

PR #454, opened 2019-10-03
Add a constructor from a flatbuffer vector to a game-state vector

robbai:


robbai:
This is available for the DesiredRotation (from a flatbuffer Rotator), but strangely not for the DesiredVector3

PR #463, opened 2020-01-01
Solved the mystery of the disappearing parenthesis

ViliamVadocz:
The opening brackets shall have their closing brethren once more.


tarehart:
Thanks for fixing. Where does this problem show up in the bot making experience? Just as a typo in the config file comment, or is it causing problems at runtime?

ViliamVadocz:
just a typo in the config file comment

PR #466, opened 2020-01-21
Deleted check for field goal.

skyborgff:
This is not needed anymore as we validate that all cars are in packet.
This allows us to load custom maps without goals.


skyborgff:
> I think all the is_valid_field_info functions are unused now, could you please clean those up?
Using "search in path" i confirmed it isn't used anywhere.

PR #473, opened 2020-02-01
Adding more info to RLBotDotNet nuspec

TheBlocks:


TheBlocks:
There was a bug with conflicting tags in the nuspec before closing the issue. Pushed a new commit without the bug.

PR #477, opened 2020-02-06
Add an exception for when Rocket League is not running after the game interface has already been initialised

ViliamVadocz:
Rewrote the `connect_to_game` method and added an `is_rocket_league_running` method to catch when Rocket League is not running and we had started once already.

My motivation for this was that I was getting another error which made no sense when really what was wrong was the situation described above.

I also generalised the gitignore to ignore `.vscode/` folders anywhere (not just in src/main/cpp/RLBotInterface).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -462,7 +462,7 @@ ASALocalRun/
 src/main/cs/RLBotControllerSample/CopyFilesFromSrc.bat
 
 # VSCode IDE files.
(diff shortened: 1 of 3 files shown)


ViliamVadocz:
I don't know why the already merged commits (63c323e, 1cd83f6, 6028cae) are showing up...

PR #478, opened 2020-02-14
 Adding Psyonix bot presets. 

RLMarvin:
This allows Psyonix bot types to get their original names and appearances.This also keeps the distinction between different skill levels by Inserting "Pro" or "Rookie" Before the name preset (for example, Hound with a skill level of 0.0 will become "Rookie Hound")
I've tested it on RLBotGUI and the old GUI, and it seems to work fine.

### DIFF
--- a/src/main/python/rlbot/setup_manager.py
+++ b/src/main/python/rlbot/setup_manager.py
@@ -12,7 +12,6 @@ from pathlib import Path
 from typing import List, Optional, Dict
 from urllib.parse import ParseResult as URL
 
-import psutil
 from rlbot.utils.structures import game_data_struct
 
 from rlbot import gateway_util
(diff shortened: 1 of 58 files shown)


src/main/python/rlbot/setup_manager.py
```diff
@@ -242,6 +243,10 @@ def load_match_config(self, match_config: MatchConfig, bot_config_overrides={}):
         self.names = [bot.name for bot in match_config.player_configs]
         self.teams = [bot.team for bot in match_config.player_configs]
 
+        for player in match_config.player_configs:
+            if player.bot and not player.rlbot_controlled:
+                set_random_psyonix_bot_preset(player)
+
         bundles = [bot_config_overrides[index] if index in bot_config_overrides else
```
RLMarvin: This line seems to be designed for overriding configs which might work for those cased. Not sure if it has been used before.

PR #482, opened 2020-02-19
Python and Rust Hiveminds

ViliamVadocz:
Explanation for how it works is on my wiki: https://github.com/ViliamVadocz/Hivemind/wiki

Plans for future:

- Add other languages
- Make hot-reloading work
- Add more in-depth explanations to the wiki

Known issue:

- If you start a match without stopping first the original helper process will not be killed and you will get ghost inputs. This can be spotted by looking at the percentages (200%, etc.)

This issue is resolved with my [PR to RLBotGUI](https://github.com/RLBot/RLBotGUI/pull/64).

---

The rust portion is in this PR: https://github.com/RLBot/rlbot-rust/pull/28

### DIFF
--- a/src/main/python/rlbot/agents/hivemind/drone_agent.py
+++ b/src/main/python/rlbot/agents/hivemind/drone_agent.py
@@ -1,0 +1,38 @@
+import os
+
+from rlbot.agents.base_independent_agent import BaseIndependentAgent
+from rlbot.botmanager.helper_process_request import HelperProcessRequest
+
+class DroneAgent(BaseIndependentAgent):
+    # Relative path to the hivemind helperprocess python file.
+    hive_path = None
+    # Bots with the same key will be part of the same hivemind.
+    hive_key = None
+    # Name of your hivemind that shows up in the console.
+    hive_name = None
+
+    def __init__(self, name, team, index):
+        super().__init__(name, team, index)
+        if self.hive_path is None:
+            raise NotImplementedError('You need to specify a path to the hivemind file.')
+        if self.hive_key is None:
+            raise NotImplementedError('You need to specify a key for your hivemind.')
+        if self.hive_name is None:
+            raise NotImplementedError('You need to specify a name for your hivemind.')
+    
+    def run_independently(self, terminate_request_event):
+        pass
+
+    def get_helper_process_request(self) -> HelperProcessRequest:
+        if os.path.isfile(self.hive_path):
+            # Appends team to key so that each team has its own hivemind.
+            key = f'{self.hive_key}{self.team}'
+            
+            # Creates request for helper process.
+            options = {
+                'name': self.hive_name
+            }
+
+            return HelperProcessRequest(self.hive_path, key, options=options)
+
+        raise FileNotFoundError('Could not find file: {self.hive_path}')
--- a/src/main/python/rlbot/agents/hivemind/drone_agent.py
+++ b/src/main/python/rlbot/agents/hivemind/drone_agent.py
@@ -35,4 +35,4 @@ class DroneAgent(BaseIndependentAgent):
 
             return HelperProcessRequest(self.hive_path, key, options=options)
 
-        raise FileNotFoundError('Could not find file: {self.hive_path}')
+        raise FileNotFoundError(f'Could not find file: {self.hive_path}')
--- a/src/main/python/rlbot/agents/hivemind/drone_agent.py
+++ b/src/main/python/rlbot/agents/hivemind/drone_agent.py
@@ -3,6 +3,7 @@ import os
 from rlbot.agents.base_independent_agent import BaseIndependentAgent
 from rlbot.botmanager.helper_process_request import HelperProcessRequest
 
+
 class DroneAgent(BaseIndependentAgent):
     # Path to the hivemind helperprocess python file.
     hive_path = None
@@ -19,20 +20,21 @@ class DroneAgent(BaseIndependentAgent):
             raise NotImplementedError('You need to specify a key for your hivemind.')
         if self.hive_name is None:
             raise NotImplementedError('You need to specify a name for your hivemind.')
-    
+
     def run_independently(self, terminate_request_event):
         pass
 
     def get_helper_process_request(self) -> HelperProcessRequest:
-        if os.path.isfile(self.hive_path):
-            # Appends team to key so that each team has its own hivemind.
-            key = f'{self.hive_key}{self.team}'
-            
-            # Creates request for helper process.
-            options = {
-                'name': self.hive_name
-            }
+        if not os.path.isfile(self.hive_path):
+            raise FileNotFoundError(f'Could not find file: {self.hive_path}')
 
-            return HelperProcessRequest(self.hive_path, key, options=options)
+        # Appends hive_path to key so that hiveminds in different places don't compete.
+        # Appends team to key so that each team has its own hivemind.
+        key = f'{self.hive_path}{self.hive_key}{self.team}'
 
-        raise FileNotFoundError(f'Could not find file: {self.hive_path}')
+        # Creates request for helper process.
+        options = {
+            'name': self.hive_name
+        }
+
+        return HelperProcessRequest(self.hive_path, key, options=options)
--- a/src/main/python/rlbot/agents/hivemind/drone_agent.py
+++ b/src/main/python/rlbot/agents/hivemind/drone_agent.py
@@ -4,7 +4,7 @@ from rlbot.agents.base_independent_agent import BaseIndependentAgent
 from rlbot.botmanager.helper_process_request import HelperProcessRequest
 
 class DroneAgent(BaseIndependentAgent):
-    # Relative path to the hivemind helperprocess python file.
+    # Path to the hivemind helperprocess python file.
     hive_path = None
     # Bots with the same key will be part of the same hivemind.
     hive_key = None
--- a/src/main/python/rlbot/agents/hivemind/python_hivemind.py
+++ b/src/main/python/rlbot/agents/hivemind/python_hivemind.py
@@ -8,6 +8,7 @@ from rlbot.botmanager.bot_helper_process import BotHelperProcess
 from rlbot.messages.flat import MatchSettings
 from rlbot.utils.game_state_util import GameState
 from rlbot.utils.logging_utils import get_logger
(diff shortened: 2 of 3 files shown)


tarehart:
Thanks for the hive mind work! I'll look at this soon, I'm waiting for a chance to really focus on it

src/main/python/rlbot/agents/hivemind/drone_agent.py
```diff
@@ -0,0 +1,38 @@
...
+            key = f'{self.hive_key}{self.team}'
+            
+            # Creates request for helper process.
+            options = {
+                'name': self.hive_name
+            }
+
+            return HelperProcessRequest(self.hive_path, key, options=options)
+
+        raise FileNotFoundError(f'Could not find file: {self.hive_path}')
```
tarehart: May want to also append the hive_path onto the key. Without that, it's possible (but unlikely) that two different botmakers could use the same key, be on the same team, and accidentally get routed to the same process.

Also, I have a small preference to have the code be like 
```
if not os.path.isfile(self.hive_path):
    raise FileNotFoundError(f'Could not find file: {self.hive_path}')
```
So that there's less indentation and therefore less effort to keep track of what's going on when reading the code. You don't have to change it if this is your preferred style though.

src/main/python/rlbot/agents/hivemind/python_hivemind.py
```diff
@@ -0,0 +1,157 @@
...
+        # Sets up the logger. The string is the name of your hivemind.
+        # This is configured inside your config file under hivemind_name.
+        self.logger = get_logger(options['name'])
+
+        # Give a warning if exec_path was given.
+        if 'exec_path' in options:
+            self.logger.warning(
+                """An exec_path was given, but this is a PythonHivemind, and so it is ignored.
+                Try subclassing SubprocessHivemind if you want to use an executable."""
+            )
```
tarehart: I love your good error messages!

src/main/python/rlbot/agents/hivemind/python_hivemind.py
```diff
@@ -0,0 +1,157 @@
...
+        # which requested this hivemind with the same key.
+        self.drone_indices = set()
+
+    def try_receive_agent_metadata(self):
+        """Adds all drones with the correct key to our set of running indices."""
+        while not self.metadata_queue.empty():
+            # Tries to get the next agent from the queue.
+            single_agent_metadata: AgentMetadata = self.metadata_queue.get(timeout=1.0)
+            # Adds the index to the drone_indices.
+            self.drone_indices.add(single_agent_metadata.index)
```
tarehart: I'm concerned that this may not wait long enough for all the metadata to arrive. What if the first bot starts the helper process and sends metadata, then this runs before the second has sent metadata? Then metadata_queue.get() will return the first metadata instantly, and the loop will exit instantly because the queue is now empty. Have you ever seen any behavior like that where not all indices get received?

The scratch manager seems to have a risky implementation also, but it sleeps a whole second before calling this method, then waits for 0.1 seconds each loop for additional metadata to arrive.

src/main/python/rlbot/agents/hivemind/python_hivemind.py
```diff
@@ -0,0 +1,157 @@
...
+        # Creating ball prediction and field info objects to later update in wrapper methods.
+        self._ball_prediction = BallPrediction()
+        self._field_info = FieldInfoPacket()
+        self.game_interface.update_field_info_packet(self._field_info)
+        # Wrapper for renderer.
+        self.renderer: RenderingManager = self.game_interface.renderer
+
+        # Create packet object.
+        packet = GameTickPacket()
+        self.game_interface.update_live_data_packet(packet)
```
tarehart: You should probably use self.game_interface.fresh_live_data_packet instead to minimize latency

src/main/python/rlbot/agents/hivemind/python_hivemind.py
```diff
@@ -0,0 +1,157 @@
...
+        while True:
+            previous_packet_time = packet.game_info.seconds_elapsed
+
+            # Updating the packet.
+            self.game_interface.update_live_data_packet(packet)
+
+            # Checking if packet is old. If yes, sleep and skip running the code.
+            if previous_packet_time == packet.game_info.seconds_elapsed:
+                time.sleep(0.003)
+                continue
```
tarehart: If you use fresh_live_data_packet here too, you won't need the time checking code since that all happens internally.

src/main/python/rlbot/agents/hivemind/subprocess_hivemind.py
```diff
@@ -0,0 +1,82 @@
...
+        self.logger.info("Collecting drones; give me a moment.")
+        self.try_receive_agent_metadata()
+        self.logger.info("Starting subprocess!")
+
+        if not os.path.isfile(self.exec_path): 
+            raise FileNotFoundError(f"Could not find file: {self.exec_path}")
+
+        # Starts the process.
+        # The drone indices are all the numbers after --drone-indices.
+        # example.exe --rlbot-verson 1.35.5 --rlbot-dll-directory some/directory/dll/ --drone-indices 0 1 2 3 4
```
tarehart: Having spaces between the drone indices like that might be awkward for a lot of command line parsers. I bet something like --drone-indices 0,1,2,3 would be easier.

src/main/python/rlbot/agents/hivemind/subprocess_hivemind.py
```diff
@@ -0,0 +1,82 @@
...
+import queue
+import os
+
+import rlbot
+from rlbot.botmanager.agent_metadata import AgentMetadata
+from rlbot.botmanager.bot_helper_process import BotHelperProcess
+from rlbot.utils.logging_utils import get_logger
+from rlbot.utils.structures.game_interface import get_dll_directory
+
+class SubprocessHivemind(BotHelperProcess):
```
tarehart: It looks like this class supports executable hive minds which prefer to have all indices passed to them as a command line argument, rather than launching with none and waiting for signals like ExecutableWithSocketAgent. That seems like a valid use case, and probably the thing you need to support the rust way of doing things right?

src/main/python/rlbot/agents/hivemind/subprocess_hivemind.py
```diff
@@ -0,0 +1,82 @@
...
+                '--drone-indices'
+            ] + [
+                str(index) for index in self.drone_indices
+            ]
+        )
+
+        # Block until we are asked to terminate, and then do so.
+        # terminate_request_event.wait()
+        # process.send_signal(SIGTERM)
+        # process.wait()
```
tarehart: What's going on down here?

src/main/python/rlbot/agents/hivemind/subprocess_hivemind.py
```diff
@@ -0,0 +1,82 @@
...
+        self.drone_indices = set()
+
+    def try_receive_agent_metadata(self):
+        """Adds all drones with the correct key to our set of running indices."""
+        while not self.metadata_queue.empty():
+            # Tries to get the next agent from the queue.
+            single_agent_metadata: AgentMetadata = self.metadata_queue.get(
+                timeout=1.0)
+            # Adds the index to the drone_indices.
+            self.drone_indices.add(single_agent_metadata.index)
```
tarehart: Same concern about this not waiting long enough.

src/main/python/rlbot/agents/hivemind/subprocess_hivemind.py
ViliamVadocz: Pretty much. I have not looked at languages that use sockets yet. Rust bots receive the index as a command line argument so I did it the same way here.

src/main/python/rlbot/agents/hivemind/python_hivemind.py
ViliamVadocz: Your compliment made me realise this is actually outdated!
I no longer pass the `exec_path` as an option. Instead I set it as a class attribute when inheriting from `SubprocessHivemind`. Fixing this ASAP.

src/main/python/rlbot/agents/hivemind/subprocess_hivemind.py
ViliamVadocz: Done :)

src/main/python/rlbot/agents/hivemind/subprocess_hivemind.py
ViliamVadocz: I copied that from `BaseSubprocessAgent` and I think it waits for a `termination_request_event` and then kills the process. Seems to work fine without it.

src/main/python/rlbot/agents/hivemind/python_hivemind.py
ViliamVadocz: I have not had a problem with this. I have run up to 8 bots in a hivemind, beyond that I get errors like this:
`INFO:rlbot[process_configuration.py:67 -  configure_processes() ] psutil.NoSuchProcess no process found with pid 17504`

Increasing the timeout did nothing to remedy the problem.

src/main/python/rlbot/agents/hivemind/subprocess_hivemind.py
ViliamVadocz: Have not had a problem with this yet.

src/main/python/rlbot/agents/hivemind/python_hivemind.py
ViliamVadocz: That simplifies things nicely!

src/main/python/rlbot/agents/hivemind/drone_agent.py
ViliamVadocz: Done.

src/main/python/rlbot/agents/hivemind/python_hivemind.py
tarehart: OK! We can improve the situation incrementally, I'm OK with pushing this as it currently is since it doesn't impact the class design.

src/main/python/rlbot/agents/hivemind/subprocess_hivemind.py
ViliamVadocz: Figured out what it does and how events work. It now successfully kills the process when a `quit_event` is set.

PR #483, opened 2020-02-21
Add the ability to set a separate set of TASystemSettings.ini for RLBot

L0laapk3:
This PR adds the ability for RLBot to modify TASystemSettings.ini.

By default, it does not do anything, but if the user creates a TASystemSettings.RLBot.ini file, it will back up the original TASystemSettings.ini, and merge the two settings into one file.

### DIFF
--- a/src/main/python/rlbot/setup_manager.py
+++ b/src/main/python/rlbot/setup_manager.py
@@ -11,6 +11,8 @@ from datetime import datetime, timedelta
 from pathlib import Path
 from typing import List, Optional, Dict
 from urllib.parse import ParseResult as URL
+import configparser
+import atexit
 
 from rlbot.utils.structures import game_data_struct
 
@@ -44,6 +46,8 @@ if platform.system() == 'Windows':
 # By default, look for rlbot.cfg in the current working directory.
 DEFAULT_RLBOT_CONFIG_LOCATION = os.path.realpath('./rlbot.cfg')
 RLBOT_CONFIGURATION_HEADER = 'RLBot Configuration'
+
+TAGAME_CONFIG_LOCATION = os.path.expanduser("~/Documents/My Games/Rocket League/TAGame/Config")
 
 
 class ROCKET_LEAGUE_PROCESS_INFO:
@@ -165,6 +169,7 @@ class SetupManager:
             self.logger.info("Will not start Rocket League because this is configured as a client!")
         # Launch the game if it is not running.
         elif not self.is_rocket_league_running(port):
+            self.mergeTASystemSettings()
             self.launch_rocket_league(port=port)
 
         try:
@@ -425,6 +430,8 @@ class SetupManager:
         time.sleep(2)  # Wait a moment. If we look too soon, we might see a valid packet from previous game.
         self.game_interface.wait_until_valid_packet()
         self.logger.info("Match has started")
+        
+        self.cleanUpTASystemSettings()
 
     def infinite_loop(self):
         instructions = "Press 'r' to reload all agents, or 'q' to exit"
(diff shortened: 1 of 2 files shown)


L0laapk3:
One example use case enabled by this change is to cap rocket league's FPS only when RLBot launches: this is very useful on high refresh rate (144hz+) monitors. Rocket leagues FPS is uncapped by setting `AllowPerFrameSleep=False`. It can then be capped again only for RLBot by creating a new two line file TASystemSettings.RLBot.ini file like so:
```
[SystemSettings]
AllowPerFrameSleep=True
```

src/main/python/rlbot/setup_manager.py
```diff
@@ -11,6 +11,8 @@
 from pathlib import Path
 from typing import List, Optional, Dict
 from urllib.parse import ParseResult as URL
+import configparser
+import atexit
```
tarehart: Do these require special packages to be installed from pypi, or are they built in to python? If it's the former, then there's some additional work to be done in setup.py.

src/main/python/rlbot/setup_manager.py
```diff
@@ -45,6 +47,8 @@
 DEFAULT_RLBOT_CONFIG_LOCATION = os.path.realpath('./rlbot.cfg')
 RLBOT_CONFIGURATION_HEADER = 'RLBot Configuration'
 
+TAGAME_CONFIG_LOCATION = os.path.expanduser("~/Documents/My Games/Rocket League/TAGame/Config")
```
tarehart: This path won't work for everyone, for example it's known to be a little different if people have OneDrive enabled, and I could imagine it also being different based on locale. Not a big issue since this is a fringe feature, but I'll be reluctant to advertise it in our wikis etc unless it becomes more robust.

src/main/python/rlbot/setup_manager.py
```diff
@@ -575,6 +582,53 @@ def kill_matchcomms_server(self):
             self.matchcomms_server.close()
             self.matchcomms_server = None
 
+    def mergeTASystemSettings(self):
```
tarehart: It seems like you could move this logic into its own file pretty easily, is that true? I wish setup_manager.py was a smaller file.

src/main/python/rlbot/setup_manager.py
L0laapk3: These are built in packages :]

src/main/python/rlbot/setup_manager.py
L0laapk3: Any suggestions to make it more robust?

src/main/python/rlbot/setup_manager.py
L0laapk3: Done this as of 43ab9b1

src/main/python/rlbot/setup_manager.py
tarehart: Well, my location is C:\Users\tareh\OneDrive\Documents\My Games\Rocket League, for some stupid reason. Also this works for windows only, obviously.

I wouldn't worry about it much, I think we can just let this feature be mostly undocumented.

L0laapk3: Do you have any idea of the reason why it is that for you?

Since there used to be a mechanism in RLBot already that edited the .ini files, did it also have to figure out the correct location there? If that is the case, maybe this code can be improved by using the more robust logic to find that folder.

tarehart: It was real simple and real hacky and never supported Linux or Mac:
```
def find_rlbot_ini():
    for games_folder in ['~/Documents/My Games', '~/OneDrive/Documents/My Games']:
        path_to_test = os.path.expanduser(f'{games_folder}/Rocket League/TAGame/Config/TARLBot.ini')
        if os.access(path_to_test, os.F_OK):
            return path_to_test
    return None
```

PR #494, opened 2020-03-28
Loadout generator feature.

tarehart:
Gives the ability to generate your bot's loadout with a python script. This lets you choose different loadouts based on player index, randomize parts of your loadout, etc. The script is executed at the beginning of every match.

To use this feature, put a line like `loadout_generator = loadout_generator_example.py` in your bot cfg file, in the [Locations] section. Then create a python file like this:
```python
import random
from pathlib import Path

from rlbot.agents.base_loadout_generator import BaseLoadoutGenerator
from rlbot.matchconfig.loadout_config import LoadoutConfig


class SampleLoadoutGenerator(BaseLoadoutGenerator):
    def generate_loadout(self, player_index: int, team: int) -> LoadoutConfig:

        # You could start with a loadout based on a cfg file in the same directory as this generator
        loadout = self.load_cfg_file(Path('appearance'), team)

        # Or you could start from scratch like this
        # loadout = LoadoutConfig()

        if player_index == 0:
            loadout.antenna_id = 287  # use a Psyonix flag if you're the first player

        loadout.team_color_id = player_index  # Different primary color depending on player index
        loadout.paint_config.wheels_paint_id = random.choice([1, 2, 4, 5])  # Random wheel color

        return loadout
```

### DIFF
--- a/src/main/python/rlbot/agents/base_agent.py
+++ b/src/main/python/rlbot/agents/base_agent.py
@@ -18,6 +18,7 @@ BOT_CONFIG_MODULE_HEADER = 'Locations'
 BOT_CONFIG_AGENT_HEADER = 'Bot Parameters'
 BOT_CONFIG_DETAILS_HEADER = 'Details'
 PYTHON_FILE_KEY = 'python_file'
+LOADOUT_GENERATOR_FILE_KEY = 'loadout_generator'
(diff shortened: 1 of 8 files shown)


tarehart:
Thanks!

PR #496, opened 2020-05-12
Hivemind support for Java (and Kotlin)

NicEastvillage:
This PR adds hivemind support for Java bots without breaking any backwards compatibility. Hiveminds are created very similarly to normal bots, except they they use the `HivemindManager` and the `HiveSocketServer` class:

```java
HivemindManager hivemindManager = new HivemindManager(team -> new MyHivemind(team));
HiveSocketServer server = new HiveSocketServer(port, hivemindManager);
server.start();
```
A tested example bot can be found at <https://github.com/NicEastvillage/RLBotJavaHivemindTemplate>

### DIFF
--- a/src/main/java/rlbot/manager/HivemindManager.java
+++ b/src/main/java/rlbot/manager/HivemindManager.java
@@ -7,11 +7,10 @@ import rlbot.cppinterop.RLBotDll;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
 
-public class HivemindManager extends BotManager {
+public class HivemindManager extends BaseBotManager {
 
     private final HivemindProcess[] hivemindProcesses = new HivemindProcess[2];
 
@@ -43,6 +42,7 @@ public class HivemindManager extends BotManager {
                 if (latestPacket != null) {
 
                     Set<Integer> droneIndexes = hivemindProcesses[team].getDroneIndexes();
+                    if (droneIndexes.isEmpty()) continue;
                     Map<Integer, ControllerState> hivemindOutput = hivemind.processInput(droneIndexes, latestPacket);
 
                     if (droneIndexes.size() > hivemindOutput.size())
@@ -69,6 +69,7 @@ public class HivemindManager extends BotManager {
         }
     }
 
+    @Override
     public Set<Integer> getRunningBotIndices() {
         Set<Integer> blueIndexes = hivemindProcesses[0].getDroneIndexes();
         Set<Integer> orangeIndexes = hivemindProcesses[1].getDroneIndexes();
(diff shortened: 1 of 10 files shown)


ViliamVadocz:
👀

NicEastvillage:
I fixed a few bugs and created an example hivemind bot.

src/main/java/rlbot/manager/HivemindManager.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.function.Function;
+
+/**
+ * This class is a BotManager for hiveminds. It runs the main logic loops and retrieves GameTickPackets for
+ * the hiveminds.
+ */
+public class HivemindManager extends BaseBotManager {
+
+    // There are only every two hiveminds of this bot, once for each team
+    private final HivemindProcess[] hivemindProcesses = new HivemindProcess[2];
```
tarehart, line 20: It's theoretically possible that a family (like the Relief family) could have two separate hive implementations (e.g. AdversityHive and AirBudHive) playing on the same team. Possible to support that?

NicEastvillage: First off, AdversityHive sounds scary af.

Secondly, yes-ish. It is currently not supported by the HivemindManager, but the user can still do it right now. They simply have to subdivide their drones based on the names. And they have complete control over how they group the bots, so they could even do pairs of AdversityBots and AirBots. The user has to create that intermediate layer themselves though.

Hm, let me just think about how to support it from the HivemindManager. The python implementation has something called "hivemind keys", and maybe that's a concept that can be reused.

ViliamVadocz: There's no point in combining two different types of bot since the will all be part of the hivemind and therefore the only difference would be appearance. (Unless you make the hivemind treat bots differently based on name)

NicEastvillage: The point is, that a family like Relief shares codebase and is thus forced to be the same hive right now. If you want to make multiple hives based on the same codebase, you are in trouble.

tarehart: Yep! This change is awesome though. I don't want to increase the complexity too much with my edge case, so feel free to say "no it's getting too messy," I would still approve.

NicEastvillage: I managed to make it work! I put it on a different branch so we can discuss it separately. See #497.

src/main/java/rlbot/manager/HivemindManager.java
```diff
@@ -0,0 +1,122 @@
...
+     * Will not include indices from humans, bots in other languages, or bots in other java processes.
+     *
+     * This may be useful for driving a basic status display.
+     */
+    public Set<Integer> getRunningBotIndicesForOrange() {
+        if (hivemindProcesses[1] != null && hivemindProcesses[1].isRunning()) {
+            return hivemindProcesses[1].getDroneIndices();
+        }
+        return new HashSet<>();
+    }
```
ViliamVadocz, line 111: What about combining these two functions into one with the team as a parameter?

ViliamVadocz: You did this in your other PR #497

PR #498, opened 2020-05-16
r and q commands for mac (and linux?)

naturecodevoid:
This allows mac users (and hopefully linux users) to run the q and r commands in the console. Previously, this was windows only.


tarehart:
This only happens if the console has focus, I hope?

naturecodevoid:
I didn't think of testing that, but I'd guess it only works when focusing on the terminal. I will test it soon.

naturecodevoid:
I have tested this and it will only fire the command if the terminal window is open (at least on mac)

tarehart:
OK sounds good. Thanks for this!

PR #499, opened 2020-05-18
Replaced  `multiprocessing.Queue` with safe `Queue`

naturecodevoid:
See `src/main/python/rlbot/utils/queue.py`


DomNomNom:
Thanks for contributing and keeping up OS support. I have two concerns here:

- Where does the request to support `.qsize()` come from? I haven't seen been able to find any use of it in the repo. Adding the extra locking would be a non-zero performance cost and I'd like that to be weighed up by some other gain. Whenever I've interacted with queues using `.empty()` instead was always sufficient for me and that method [does not have the same disclaimer](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue.empty).

- Python's `queue.Queue` (rather than `multiprocessing.Queue`) [does not have the same disclaimer](https://docs.python.org/3/library/queue.html#queue.SimpleQueue.qsize). What are the benefits/tradeoffs of using the more complex queue there? Would it make sense to only replace uses of `multiprocessing.Queue`?

naturecodevoid:
> * Where does the request to support `.qsize()` come from? I haven't seen been able to find any use of it in the repo. Adding the extra locking would be a non-zero performance cost and I'd like that to be weighed up by some other gain. Whenever I've interacted with queues using `.empty()` instead was always sufficient for me and that method [does not have the same disclaimer](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue.empty).

I've been using qsize in a bot and though it would be good to add a modified version to the framework to ensure OS X works. Not sure of any other gains, though.

> * Python's `queue.Queue` (rather than `multiprocessing.Queue`) [does not have the same disclaimer](https://docs.python.org/3/library/queue.html#queue.SimpleQueue.qsize). What are the benefits/tradeoffs of using the more complex queue there? Would it make sense to only replace uses of `multiprocessing.Queue`?

I'm fine with using queue.Queue, this just seemed like a good solution. I don't think there are many benefits besides that my Queue should work with the rest of muiltiprocessing.

src/main/python/rlbot/setup_manager.py
```diff
@@ -14,6 +14,9 @@
 
 from rlbot.utils.structures import game_data_struct
 
+from rlbot.utils.queue import Queue as _queue
+mp.Queue = _queue
```
DomNomNom, line 18: I'd recommend against modifying module objects and instead changing uses of `mp.Queue` to `Queue`

naturecodevoid: Ok. Sorry, I won't be able to update this until monday

naturecodevoid:
Should be good to merge now

src/main/python/rlbot/botmanager/helper_process_manager.py
```diff
@@ -7,6 +7,8 @@
 
 from rlbot.botmanager.bot_helper_process import BotHelperProcess
 
+from rlbot.utils.queue import Queue as _queue
+mp.Queue = _queue
```
ViliamVadocz, line 11: Following previous commit, is this correct?

DomNomNom: This should follow the same pattern as in setup_manager.py
https://github.com/RLBot/RLBot/pull/499/commits/4600e0fd58a55c37aefe2d5bbd92203308f63ae5#

DomNomNom:
Note that the difference between queue.Queue and mp.Queue is that mp.Queue serializes and deserializes any object passed through it and queue.Queue shares memory directly. https://stackoverflow.com/a/925241

VirxEC:
Stale

PR #501, opened 2020-05-25
Support for scripts

tarehart:
Three things going on here:

**Changing directory to be more forgiving on relative paths**
Often when people are developing a python bot, they will launch it using run.py in that local directory, and start to make assumptions that they can rely on relative paths from that working directory. But then things break when running from RLBotGUI, for example. Using chdir to make things behave consistently as expected.

**Doing less manipulation of modules when loading bot classes**
A long time ago we were having issues where certain bots didn't like loading at the same time as other bots because they had conflicting python module dependencies. To work around that, I tried removing the modules from the list as soon as a bot was done loading, but this causes havoc with various libraries, including connexion which I'm currently working with. I believe this technique may not be needed anymore since the python bots are now all in separate processes except for certain edge cases in the old GUI. I tested with a large number of bots from the bot pack and did not see any issues.

**Support for running scripts during a game**
Sometimes you want to run a script during the game which does not have a car associated with it, e.g. setting zero gravity for space jam, or gathering data for a twitch overlay. I set things up so that these scripts can be configured fairly easily, using a similar cfg format / directory scanning technique. Coming soon to the GUI:

image

### DIFF
--- a/rlbot.cfg
+++ b/rlbot.cfg
@@ -129,3 +129,6 @@ participant_loadout_config_8 = None
 participant_loadout_config_9 = None
 
 
+[Scripts]
+script_config_0=src/test/python/agents/script/sample_script.cfg
+script_config_1=src/test/python/agents/script/sample_script.cfg
--- a/src/main/python/rlbot/parsing/bot_config_bundle.py
+++ b/src/main/python/rlbot/parsing/bot_config_bundle.py
@@ -1,13 +1,16 @@
 import configparser
+import itertools
 import os
 from pathlib import Path
 
 from rlbot.agents.base_agent import BaseAgent, BOT_CONFIG_MODULE_HEADER, BOT_NAME_KEY, LOOKS_CONFIG_KEY, \
     PYTHON_FILE_KEY, LOGO_FILE_KEY, SUPPORTS_EARLY_START_KEY, LOADOUT_GENERATOR_FILE_KEY
 from rlbot.agents.base_loadout_generator import BaseLoadoutGenerator
+from rlbot.agents.base_script import SCRIPT_FILE_KEY, BaseScript
+from rlbot.agents.rlbot_runnable import RLBotRunnable
 from rlbot.matchconfig.loadout_config import LoadoutConfig
 from rlbot.parsing.agent_config_parser import create_looks_configurations, PARTICIPANT_CONFIGURATION_HEADER, \
-    PARTICIPANT_CONFIG_KEY, load_bot_appearance
+    PARTICIPANT_CONFIG_KEY, load_bot_appearance, SCRIPT_CONFIGURATION_HEADER, SCRIPT_CONFIG_KEY
 from rlbot.parsing.custom_config import ConfigObject
 from rlbot.utils.class_importer import import_class_with_base
 from rlbot.utils.logging_utils import get_logger
@@ -15,18 +18,15 @@ from rlbot.utils.logging_utils import get_logger
 logger = get_logger('rlbot')
 
 
-class BotConfigBundle:
+class RunnableConfigBundle:
     def __init__(self, config_directory, config_obj: ConfigObject, config_file_name: str = None):
         self.config_directory = config_directory
         self.config_file_name = config_file_name
         self.config_path = os.path.join(self.config_directory, self.config_file_name)
         self.config_obj = config_obj
-        self.base_agent_config = BaseAgent.base_create_agent_configurations()
+        self.base_agent_config = RLBotRunnable.base_create_agent_configurations()
         self.base_agent_config.parse_file(self.config_obj, config_directory=config_directory)
         self.name = config_obj.get(BOT_CONFIG_MODULE_HEADER, BOT_NAME_KEY)
-        self.looks_path = self.get_absolute_path(BOT_CONFIG_MODULE_HEADER, LOOKS_CONFIG_KEY)
-        self.python_file = self.get_absolute_path(BOT_CONFIG_MODULE_HEADER, PYTHON_FILE_KEY)
-        self.loadout_generator_file = self.get_absolute_path(BOT_CONFIG_MODULE_HEADER, LOADOUT_GENERATOR_FILE_KEY)
         self.supports_early_start = self.base_agent_config.get(BOT_CONFIG_MODULE_HEADER, SUPPORTS_EARLY_START_KEY)
 
     def get_logo_file(self):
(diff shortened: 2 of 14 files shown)


ViliamVadocz:
> **Support for running scripts during a game**
> e.g. setting zero gravity for space jam

Could these scripts be included in the bot pack or maybe something separate like a script pack?

tarehart:
Yep it's possible to do either of those and I think that's a good idea 👍

Darxeal:
Nice work!
One thing that would be nice to have is a base class for scripts that want to run something every frame, similiar to `get_output`. I'm not sure if that can be done only with something like a while loop with time.sleep, or if there's a better way to do it

tarehart:
Thanks! I think a loop similar to get_output will be a common use case, but specific implementations will vary.
- Does the script want to wait for a fresh packet or not?
- Do they need custom tick rate logic?
- How do they wish to exit the loop?

And then if we ever change any of that in the framework, we'd have to worry about whether it breaks anyone's script. I'd rather just let people write custom code for their loop, it's not very difficult: https://github.com/tarehart/RLBotTwitchBroker/blob/master/tst/boost_monkey.py#L62-L79

rlbot.cfg
```diff
@@ -129,3 +129,6 @@ participant_loadout_config_8 = None
 participant_loadout_config_9 = None
 
 
+[Scripts]
+script_config_0=src/test/python/agents/script/sample_script.cfg
+script_config_1=src/test/python/agents/script/sample_script.cfg
```
DomNomNom, line 134: How come the sample script is listed twice in the default cfg?

src/main/python/rlbot/parsing/bot_config_bundle.py
```diff
@@ -118,3 +149,19 @@ def get_bot_config_bundles(num_participants, config: ConfigObject, config_locati
...
     return config_bundles
+
+
+def get_script_config_bundles(match_config: ConfigObject, config_location):
+    """
+    Adds all the script config bundles found in the match config file.
+    Attempts to read indexed config keys until it hits an empty one, up to a maximum of 100
+    """
+    config_bundles = []
+    for i in range(100):
```
DomNomNom, line 160: This 100 seems a bit limiting. Would a while loop be better?
```suggestion
    while script_config_relative_path is not None:
```

src/main/python/rlbot/utils/class_importer.py
```diff
@@ -79,18 +78,11 @@ def load_external_module(python_file):
 
     dir_name = os.path.dirname(python_file)
     module_name = os.path.splitext(os.path.basename(python_file))[0]
-    keys_before = set(sys.modules.keys())
```
DomNomNom, line 82: I remember using things from this file in RLBotTraining. Have you tested whether there is an impact?
https://github.com/RLBot/RLBotTraining/blob/580e162e54f41d278319bf4f68d99dab7f3b7ed8/rlbottraining/exercise_runner.py#L136

src/test/python/agents/script/sample_script.cfg
```diff
@@ -0,0 +1,16 @@
...
+[Details]
+# These values are optional but useful metadata for helper programs
+# Name of the bot's creator/developer
+developer = tarehart
+
+# Link to github repository
+github = https://github.com/tarehart/RLBotTwitchBroker
+
+# Programming language
+language = python
```
DomNomNom, line 16: Do these support parameters similar to bots?

src/test/python/agents/script/sample_script.py
```diff
@@ -0,0 +1,20 @@
...
+            self.renderer.begin_rendering()
+            self.renderer.draw_string_2d(random.randint(200, 300), random.randint(20, 300), 1, 1,
+                                         str(packet.game_info.seconds_elapsed), self.renderer.yellow())
+            self.renderer.end_rendering()
+            sleep(5)
+
+
+print("Sample script starting...")
+script = SampleScript("myScript")
+script.run()
```
DomNomNom, line 20: Please document when things get loaded. 
I think it's dangerous to do state-changing things at import time (prevents reuse) and I'd recommend having a callback (e.g. `on_added_to_match`) instead.

DomNomNom:
Overall looks good :+1:

rlbot.cfg
tarehart: I wanted to prove to myself that it would successfully load two scripts, but was too lazy to actually make two unique test scripts.

src/main/python/rlbot/parsing/bot_config_bundle.py
tarehart: Yeah fair enough, I'll make it a while loop. Using more than 100 scripts is probably a poor idea, but if anyone ever tries, we can probably be entertained!

src/main/python/rlbot/utils/class_importer.py
tarehart: I ran a match with a bunch of bots from the bot pack, but did not test anything in RLBotTraining.  Thanks for calling this out, I'll test training.

src/test/python/agents/script/sample_script.py
tarehart: Would it be adequate for the user to have one of these?
```python
if __name__ == '__main__':
    do_whatever()
```

I'd prefer to avoid the callback pattern because that would add more complexity and constraints I think:
- The scripts must extend a base class (currently it's purely optional)
- The framework needs to do class loading for the scripts, which has historically been tricky

src/test/python/agents/script/sample_script.cfg
tarehart: Nope. With bots, I believe the framework reads the config and feeds it to the bot in some callback / lifecycle type function. Scripts get no callbacks, they're on their own. There's nothing stopping them from reading their own config file, though.

I'm inclined to leave it the way it is, I like having loose coupling for these scripts.

src/test/python/agents/script/sample_script.py
DomNomNom: - I'm note sure if the `if __name__ == '__main__'` helps here as you should only get `'__main__'` back if you're being run at top level and are expected to do things like command line flag parsing. It's possible to import it with a custom name but that doesn't appeal to me.
- I was thinking that `on_added_to_match()` could just be functions like how RLBotTraining does it with `make_default_playlist()` [[1]](https://github.com/RLBot/RLBotTraining/blob/580e162e54f41d278319bf4f68d99dab7f3b7ed8/rlbottraining/exercise_runner.py#L137) [[2]](https://github.com/RLBot/RLBotTraining/blob/580e162e54f41d278319bf4f68d99dab7f3b7ed8/rlbottraining/common_exercises/bronze_striker.py#L79)
- We seem to be doing module loading already.

src/test/python/agents/script/sample_script.cfg
DomNomNom: Cool

src/main/python/rlbot/parsing/bot_config_bundle.py
DomNomNom: I think 100+ scripts is overkill too but it would be unpleasantly surprising if it just started dropping scripts without an error. So rather than throwing an error, it's probably nicer to do it with a `while`.

src/test/python/agents/script/sample_script.py
tarehart: `if __name__ == '__main__'` is working as intended in https://github.com/tarehart/RLBotTwitchBroker/blob/master/tst/boost_monkey.py#L99-L101

I may be misunderstanding you, but to be clear the scripts are started with `subprocess.Popen([sys.executable, script_config_bundle.script_file]...`. For scripts we're not doing any module loading yet, and they get run in a 'top level' manner.

Your `make_default_playlist()` does seem nice and optional and lightweight, but I don't see it adding much value in this case since `if __name__ == '__main__'` is working as intended.

src/main/python/rlbot/utils/class_importer.py
tarehart: Training does still work when executing `rlbottraining run_module .\hello_world_training.py`, which I believe exercises the code in question.

src/test/python/agents/script/sample_script.py
DomNomNom: :+1: I missed the `subprocess.Popen()`.

PR #506, opened 2020-06-07
Calling begin_rendering and end_rendering on behalf of python bots.

tarehart:
I put Botimus, Beast from the East, SkyBot and Kamael in the game, and they're all still rendering fine.


ViliamVadocz:
Would this mean you could render without opening and closing a render group?

tarehart:
Yep!

PR #509, opened 2020-06-25
Allow Psyonix bots to be preconfigured with a loadout

azeemba:
If a PlayerConfig is a psyonix bot then it's loadout config
was always being overwrriten. However, if the loadout was
specifically set, then we should just respect it.


azeemba:
So the GUI runner in this repo actually seems to always set it to the appearance you have selected. I can't see a way to unset the appearance.

RLBotGUI does leave it as None for Psyonix bots

tarehart:
OK sounds good! GUI runner in this repo isn't really supported anymore.

PR #510, opened 2020-06-26
Allow GUI to completely shutdown RLBot including RLBot.exe

L0laapk3:
Adds an option to really shut down RLBot, including RLBot.exe. This is required for the GUI to properly terminate.


L0laapk3:
Note: Refer to https://github.com/RLBot/RLBotGUI/pull/102 for discussion about this PR.

L0laapk3:
I'm not sure what developments happened but this appears to be either in there or no longer needed as the command prompt is closed now when you close the gui, so I'm closing this

PR #513, opened 2020-07-05
Fix currentID initialization in MessageBase constructor

themaxdavitt:
Resolves compilation error "`copying variable of type 'std::atomic_uint' (aka 'atomic<unsigned int>') invokes deleted constructor`" that appears when using `clang` (for `rlbot-rust` binding generation). There's a few questions on Stack Overflow that describe the issue in greater detail if you're curious.

Haven't tested using the [instructions from the README](https://github.com/RLBot/RLBot/blob/cbe257682f70909c27c2f1c35d9b32f7bd4cf5a3/src/main/cpp/RLBotInterface/README.md) because I don't have Visual Studio installed, sorry! It's a very minor change anyways.

### DIFF
--- a/src/main/cpp/RLBotInterface/src/RLBotMessages/MessageStructs/Message.hpp
+++ b/src/main/cpp/RLBotInterface/src/RLBotMessages/MessageStructs/Message.hpp
@@ -26,7 +26,7 @@ struct MessageBase
 protected:
 	MessageBase(MessageType type, bool hasCallback) : Type(type), HasCallback(hasCallback)
 	{
-		static std::atomic_uint currentID = {0};
+		static std::atomic_uint currentID(0);
 		ID = currentID++;
 	}
 };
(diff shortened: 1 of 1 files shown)


src/main/cpp/RLBotInterface/src/RLBotMessages/MessageStructs/Message.hpp
```diff
@@ -26,7 +26,7 @@ struct MessageBase
 protected:
 	MessageBase(MessageType type, bool hasCallback) : Type(type), HasCallback(hasCallback)
 	{
-		static std::atomic_uint currentID = 0;
+		static std::atomic_uint currentID = {0};
```
azeemba, line 29: Is there a reason to not just invoke the constructor directly? I checked with godbolt and the generated code is the same but its a lot easier to read.

```cpp
static std::atomic_uint currentID(0);
```

themaxdavitt: Sorry about the huge delay. Yeah, that would be a lot better - I don't write much C++ so I didn't really think about the cleanest way to write it at the time. Thanks for the suggestion!

themaxdavitt:
@tarehart Correct. I changed the syntax slightly in a new commit based off @azeemba's suggestion (you can just squash them when merging), and you can see in [C++ Insights](https://cppinsights.io/s/870cfc64) that the same code is generated.

PR #514, opened 2020-07-06
Reset gravity value per match

azeemba:
Changing the gravity seems to last for the duration of
the RL process. So when a new match is started, we manually
reset it to the default value -650.

I didn't see a canonical place for that value, so it is directly
hardcoded.

The SetupManager only sets this value once at the start so the
ZeroG mutator script still works.

Feel free to let me know if this fix should go at a different level/in a different class.
We could also try to "revert" this value when a match ends but I think
there might be more failure cases that might need to be thought through.


tarehart:
Seems reasonable! I do have a good place for this in the closed source area though. I'll investigate that and get back to this pr afterwards. Giving myself a time limit of 7 days

tarehart:
I've now implemented a gravity reset within RLBot.exe and deployed that as rlbot 1.43.7, so I think this PR is no longer needed, but thanks for identifying the problem!

PR #526, opened 2020-09-23
Add support for the Epic Games Launcher

VirxEC:
Please view line 212, idk what I should put there

### DIFF
--- a/src/main/python/rlbot/setup_manager.py
+++ b/src/main/python/rlbot/setup_manager.py
@@ -1,6 +1,7 @@
 import multiprocessing as mp
 import os
 import sys
+import json
 
 import psutil
 import platform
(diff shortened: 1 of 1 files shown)


tarehart:
Looks good!

Probably something like this would work for the missing code:
```python
exe_and_args = [epic_exe_path] + ideal_args
_ = subprocess.Popen(exe_and_args)
```

VirxEC:
Makes sense, did that

tarehart:
Cool! Does it launch RL and start a match if you do `python runner.py` in the root directory?

I guess maybe you would need to comment out the steam code to test properly...

VirxEC:
I tested both Steam & Epic Games, both of them work; they put 2 AlwaysTowardsBallAgent bots in a heatseeker match

tarehart (approved):
Thanks Virx!

PR #527, opened 2020-09-24
Clean up code & more safe gaurds

VirxEC:
Only *.item files are now looped through
If a *.item file isn't a json file, it will just be skipped over
Message to console about whether we're launching via Steam or Epic Games
pathlib.Path is now used for the file/directory paths
Fixed bug with launching via the fall-back web browser on Linux
Cleaned up the Epic Games binary-finding code


VirxEC:
I have thoroughly tested all aspects of this code. I made sure that it was all working. There shouldn't be any bugs with it.

tarehart (approved):
Thanks for the improvements, Virx!

PR #528, opened 2020-10-04
New Epic launcher technique that allows login. Ability to pass preferences on that.

tarehart:
This will allow people to set a preference for Steam vs Epic if they have Rocket League on both. I anticipate making a change to RLBotGUI to make that more easily available.

For Epic, if they like, we will try a neat trick to harvest and incorporate a temporary key that allows login. That way they get their proper controls + camera settings. It's a little bit wild and there's a very very small chance somebody will see weird logs on an authentication server and react, so I want it to be opt-in.


tarehart:
Wow nice, approved

PR #529, opened 2020-10-27
Improving time resolution in windows to fix fresh packet delivery rate.

tarehart:
Also including a few recent changes from the closed source repo.

If anyone wants to help test this, you can check out the fresh-fix branch and run setup.bat to make your system python instance use this version of RLBot. Then you could launch RLBotGUI on your system python, e.g. like this: https://github.com/RLBot/RLBotPythonExample/blob/master/run_gui.py


VirxEC:
Looks great! Thanks for this fix 👍

Darxeal (approved):

PR #534, opened 2020-12-29
Port of RLBotDotNet from .NET Framework 4.6.1 to .NET Standard 2.0 (#534))

akaBrainlin:
In order to allow the usage of the RLBotDotNet in either .NET Framework and .NET Core, i ported the library to .NET Standard 2.0. Most of the settings of the project file and the solution file were kept intact, including GUID and compile options.

Libraries of the type .NET Standard 2.0 are fully supported to be used in .NET Framework 4.6.1+ and .NET Core 2.0+ (including the new .NET 5) projects - as stated here: https://docs.microsoft.com/en-us/dotnet/standard/net-standard

I successfully tested the library with the included update of the TestBot through RLBotGUI.

Since i'm not that kind of an expert with NuGet-commiting i kept the *.nuspec intact - this file potentially needs to be updated, too!


tarehart:
Awesome, thank you!

Regarding nuget, we should find a way to test the upgrade process for people who have existing bots. It looks like there are a few breaking changes involved, e.g. they'd need to find and replace some System.Windows.Media.Color stuff to get things compiling again. Worth a major version bump.

akaBrainlin:
> Awesome, thank you!
> 
> Regarding nuget, we should find a way to test the upgrade process for people who have existing bots. It looks like there are a few breaking changes involved, e.g. they'd need to find and replace some System.Windows.Media.Color stuff to get things compiling again. Worth a major version bump.

The struct System.Windows.Media.Color is only artifically used by saving the ARGB value, which are then parsed into the flatbuffer builder. For a existing Bot the method Color.FromRGB changes now to Color.FromARGB - this could easily handled by an Extension. The change in the using-clause of the Bot-Class is still needed though (using Color = System.Drawing.Color;).

TheBlocks (approved):
This looks great! Thanks for the PR. I want to push for .NET 5 support, so this PR helps out a ton. This definitely deserves a major version bump.

I tested it out on the TestBot and the example bot, and it works well on my setup.

PR #535, opened 2020-12-31
Standalone python bots for simpler debugging / faster startup

tarehart:
This allows a bot's python file to run standalone if they opt in to the special base class. There are two main use cases here:
- When running rlbot normally, e.g. via RLBotGUI, we can start bots as completely new processes, rather than python subprocesses. For some reason this is much faster. I had to use this technique to make PokebotTrainer feel good in the Puppetmaster 2 event.
- When developing a bot, you can just run your bot.py file directly rather than running via RLBotGUI. I think this is more intuitive for newcomers, and it's easier for IDEs to run with debugging.

It goes something like this:
```python
from rlbot.agents.base_agent import SimpleControllerState
from rlbot.agents.standalone.standalone_bot import StandaloneBot, run_bot
from rlbot.utils.structures.game_data_struct import GameTickPacket

class StandaloneTest(StandaloneBot):

    def get_output(self, game_tick_packet: GameTickPacket) -> SimpleControllerState:
        return SimpleControllerState()

if __name__ == '__main__':
    run_bot(StandaloneTest)
```

And the config file gains `supports_standalone = True` so that it looks like this:
```
[Locations]
looks_config = appearance.cfg
python_file = bot.py
supports_standalone = True
name = PythonExampleBot
maximum_tick_rate_preference = 120
```

If you run the python file with no arguments, it will open rocket league and RLBot.exe and spawn itself into a game as player index 0. If you kill the process and restart, Rocket League stays open, the car stays on the field, and the bot picks up where it left off. Hot reload still works.

What we lose:
- Ability to send explicit signal to bot telling it to reload. Used to do this with the 'r' key, generally unused now.
- Ability to send signal telling the bot to retire gracefully. However, it still retires gracefully if you end the match from within Rocket League.


src/main/python/rlbot/agents/standalone/executable_with_socket_standalone_bot.py
```diff
@@ -0,0 +1,72 @@
...
+        #     self.logger.info("Found port already available")
+        # elif self.find_process_using_port(self.get_port()):
+        #     self.logger.info("Found process already running")
+        # else:
+        launch = [self.executable_path, str(self.get_port())]
+        proc = subprocess.Popen(launch, cwd=os.path.dirname(self.executable_path))
+        process = psutil.Process(pid=proc.pid)
+        configure_process(process, get_team_cpus(self.team))
+        # TODO: handle the case where the process spans both teams; should be pinned differently
+        self.logger.info("Launched executable instance")
```
NicEastvillage, line 56: There's a TODO here. Is this something that needs fixing before release?

NicEastvillage:
> When developing a bot, you can just run your bot.py file directly rather than running via RLBotGUI. I think this is more intuitive for newcomers, and it's easier for IDEs to run with debugging.

From what I can tell, it creates a 1v0 match. 1v0 isn't really that exciting once you have kickoffs implemented, so I think it should be a 1v1 against itself. Ideally, the team sizes would be specified in the `run_bot` function.

src/main/python/rlbot/agents/standalone/executable_with_socket_standalone_bot.py
tarehart: Good catch! I wrote this part of the code half a year ago and forgot, I'll deal with it

tarehart:
Good point, 1v0 has limited usefulness. However, it's easier in the code to have only one of itself, and more intuitive I think. If they attach a debugger, they will avoid confusion from multiple bots hitting breakpoints.

How would you feel about one of itself vs a psyonix rookie? For anything more specific, e.g. bigger team sizes, I think using RLBotGUI is the best option. If you launch the GUI via pycharm, you can hit breakpoints in any python bot it spawns.

I'm a little torn about what's the most sensible default, 1v0 or 1vPsyonix. With 1v0 it's easier to jump in as a human, and maybe lower pressure for an absolute beginner (which is my main target audience).

oxrock:
Although I like the idea of a psyonix rookie being default as it's a lightweight punching bag for new developers without having to fiddle with the gui.... I think 1v0 when starting a new bot is more sensible. There's a lot of things that need to get implemented before dealing with an opponent becomes a significant factor. By the time they're ready for that, hopefully they're comfortable enough with learning how to use the rlbotgui and tailor their experience to precisely what they need going forward.

NicEastvillage:
Those are good points. When I was suggesting different team sizes, I was think of multiple of the same bot.

How about we make the function signature:
```python
def run_bot(bot: StandaloneBot, blue_team:List[TestBotType]=[TestBotType.MyBot], orange_team:List[TestBotType]=[]):
```
where TestBotType is
```python
class TestBotType(Enum):
    MyBot = 0
    PsyonixRookie = 1
    PsyonixPro = 2
    PsyonixAllStar = 3
```
So by default it starts a 1v0 match. But as soon as the dev is ready, they can change the teams to any combination of their bot and psyonix bots by changing the lists that define the team composition. They can do 1v1 against a Psyonix Rookie with `run_bot(MyCoolBot, blue_team=[MyBot], orange_team=[PsyonixRookie])` or 2v2 against itself with `run_bot(MyCoolBot, blue_team=[MyBot, MyBot], orange_team=[MyBot, MyBot])`. They can also easily test their bot on orange.

This feels like the most versatile solution. The only thing they can't do is play against other bots -- but that should be done in the GUI.

tarehart:
I agree that's more versatile, but after sleeping on it I'm leaning pretty strongly toward a pure 1v0 capability with no options. The way I'm looking at it is you need to draw the line somewhere--at what point do you tell people "ok, time to start using RLBotGUI to launch games."

I strongly don't want to run more than one instance of the user's bot, because that breaks the intuition for the user and forces me to use 'orchestration' code, i.e. setting up the bot indices, spawning additional processes. That's a big jump up in complexity and maintenance pain. I either duplicate the orchestration code and forget to maintain it, or reuse framework code and have to take more care with future changes, I don't want to do either.

As far as 1v0 or 1vPsy, I agree with oxrock that 1v0 is more sensible for a complete beginner.

I'm inclined to do 1v0 with zero options to customize. Draws a clear line, no feature creep, and launching https://github.com/RLBot/RLBotPythonExample/blob/master/run_gui.py from PyCharm is super easy and effective for people who are ready to customize.

VirxEC (approved):
I approve

NicEastvillage (approved):

PR #536, opened 2021-01-01
Adding virtual environment support for standalone bots.

tarehart:
A standalone bot can have its own virtual environment with different package versions from other bots in the same match.
For this to work, the .cfg file needs all three of these:
```
[Locations]
supports_standalone = True
use_virtual_environment = True
requirements_file = requirements.txt
```

Before the match begins, the virtual environment will be created in a `venv` folder next to the bot cfg, and anything in the requirements file will be installed to it. When the bot is run, the venv will be used instead of the interpreter RLBot is using.

This will make it much easier for people to package / vend utility libraries because they can make breaking changes and individual bots can pin themselves to different versions and still play in the same match.

This slows down the startup process a bit, but it happens *before* we load up the match, so this does not put bots in danger of missing the first kickoff.


VirxEC (approved):
**_Yes_**

tarehart:
Fixed some bugs, renamed some things, throwing it out there!

PR #537, opened 2021-01-04
Socket manager for connecting directly to RLBot.exe from Python.

tarehart:
Now that sockets is out, it's possible to bypass the dll and have Python talk directly with RLBot.exe on the socket. I've written a helper class to make this easier.

This is cool because it's the first time certain data is readily available:
- Player spectate events: When a particular player is spectated in-game, their index will be emitted
- Player input changes: When a player (bot or human) moves their controls in any way, the data is emitted
- Player stat events: Any time something happens, e.g. ball centered, hat trick, demolition, etc, an event is emitted

It can also be used to send player inputs and match starts. I haven't gotten around to implementing rendering, state setting, or quick chat, but that can be done later.

Check out the test scripts in this PR to see how it looks to an end user.

### DIFF
--- a/src/main/python/rlbot/agents/base_agent.py
+++ b/src/main/python/rlbot/agents/base_agent.py
@@ -1,10 +1,11 @@
 from typing import Optional
 from urllib.parse import ParseResult as URL
 
+from flatbuffers import Builder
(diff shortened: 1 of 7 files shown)


NicEastvillage (approved):
Looks good, though I haven't tested it. I am already brainstorming things I can do with this!

PR #538, opened 2021-01-18
Wait-until-valid should make wait for all players

azeemba:
The current logic makes sure all rlbot controlled bots
are spawned and that there are no old rlbot bots hanging around.
However, if there are only humans/psyonix
bots in the config, it doesn't wait for them to spawn.

Since the spawn-id validation makes sure there are no
unexpected bots in the packets, we just need to validate
that the length of the players is matching in the
packets.


tarehart (approved):

tarehart:
Looks good!

PR #539, opened 2021-01-23
Fix error when starting a match with a human & rename Neon_Fields to NeonFields

Darxeal:
- currently when you try to start a match with a human it errors because humans don't have a loadout config, this should fix it
- renamed Neon_Fields to NeonFields because that's what I named it in the GUI (sorry), I think it follows the naming convention better (underscores are mostly for alternative versions)

### DIFF
--- a/src/main/flatbuffers/rlbot.fbs
+++ b/src/main/flatbuffers/rlbot.fbs
@@ -636,7 +636,7 @@ enum GameMap : byte {
   RivalsArena,
   Farmstead_Night,
   SaltyShores_Night,
-  Neon_Fields
+  NeonFields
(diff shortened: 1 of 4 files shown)


Darxeal:
There's another issue, showroom also errors because `enable_lockstep` is `None`, because it hasn't been explicitly set. This can be fixed easily in RLBotGUI but I'll wait for your opinion on how to handle default values in MatchConfig

tarehart (approved):

tarehart:
I think those values should be defaulted to false in the constructor, probably. But that's an opinion from my phone. Thanks for these fixes!

PR #542, opened 2021-02-04
Fix the doc-string for StandaloneBot

ViliamVadocz:
- Add a sentence about where to find out more concerning command line arguments
- Fix typo


tarehart (approved):
Thanks!

PR #550, opened 2021-02-23
Give scripts access to matchcomms

ViliamVadocz:
- Pass the matchcomms URL as a command line argument to scripts
- Read command line arguments for matchcomms URL

Still need to do:
- Figure out if we need to close matchcomms when script terminates (and how to close it)
- Possible bug: https://discord.com/channels/348658686962696195/423167304956903428/813503008796049469
- Fix crashing during match stop (might not be directly related, but I did encounter it a lot during testing)
- Give Hiveminds access to matchcomms

### DIFF
--- a/src/main/python/rlbot/agents/base_script.py
+++ b/src/main/python/rlbot/agents/base_script.py
@@ -1,6 +1,10 @@
+import sys
 import random
+from typing import Optional
+from urllib.parse import urlparse, ParseResult as URL
 
 from rlbot.agents.rlbot_runnable import RLBotRunnable, LOCATIONS_HEADER
+from rlbot.matchcomms.client import MatchcommsClient
 from rlbot.messages.flat import MatchSettings
 from rlbot.parsing.custom_config import ConfigObject
 from rlbot.utils.game_state_util import GameState
@@ -17,6 +21,8 @@ class BaseScript(RLBotRunnable):
     It is NOT required to use this when configuring a script.
     """
 
+    matchcomms_root: Optional[URL] = None
+
     def __init__(self, name):
         super().__init__(name)
         self.logger = get_logger(name)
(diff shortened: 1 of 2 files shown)


src/main/python/rlbot/agents/base_script.py
```diff
@@ -27,6 +33,13 @@ def __init__(self, name):
         self.game_interface.load_interface()
         fake_index = random.randint(100, 10000)  # a number unlikely to conflict with bots or other scripts
         self.renderer = self.game_interface.renderer.get_rendering_manager(bot_index=fake_index, bot_team=2)
+        # Get matchcomms root if provided as a command line argument.
+        try:
+            pos = sys.argv.index("--matchcomms-url")
+            potential_url = sys.argv[pos + 1]
+            self.matchcomms_root = urlparse(potential_url)  # Not sure what to do if this is not a valid url
+        except (ValueError, IndexError):
+            pass
```
tarehart, line 42: In my opinion, it would be good to fail loud and fast here, by just letting the exception bubble up.

tarehart (commented):
Looks good! Just one comment below. Do you want to push this now, or leave it open while you continue with your TODO list?

src/main/python/rlbot/agents/base_script.py
ViliamVadocz: Those error are related to the args checking. `urlparse` doesn't raise any errors.

ViliamVadocz:
> Looks good! Just one comment below. Do you want to push this now, or leave it open while you continue with your TODO list?

I think I will try to do the first three items on my TODO list. The last one can be part of another PR.

ViliamVadocz:
## Status report

Script seem to work as expected. I managed to get a TMCP tracker script working reliably.

> Figure out if we need to close matchcomms when script terminates (and how to close it)

I think it is the reason for this error message (https://discord.com/channels/348658686962696195/423167304956903428/813909722431356978), but the match was able to restart afterwards, so it had no effect.

> Possible bug: https://discord.com/channels/348658686962696195/423167304956903428/813503008796049469

I addressed this in my latest commit c269708.

> Fix crashing during match stop (might not be directly related, but I did encounter it a lot during testing)

I am still not sure what is behind it, but I did crash even without my changes. I think the most reliable way to reproduce is to start the gui from a VSCode powershell and then restart a match while bots are using matchcomms (I used 2v2 Otters). (no scripts involved)

> Give Hiveminds access to matchcomms

I will do this in a different PR at a later date.

## Conclusion

I am happy with these changes. Feel free to merge.

tarehart (approved):
Looks good!

PR #551, opened 2021-03-18
Add team-only matchcomms (and catch an exception in process_configuration.py)

rivques:
How team-only matchcomms work:
- **_NOTE:_** This only works in Python.
- `rlbot.matchcomms.client.MatchcommsClient` automatically adds `team` to the sent packets.
- This doesn't break anything, unless you are sending a packet where `team` is the team you are not on, in which case the client will silently change it to your own team.
- If you would like to send a team-only matchcomm, add `'team_only': True` to your packet. If it is `False` or ommitted everybody can see it.
- Scripts can see all matchcomms, even team-only ones, and the ones they send are visible to everybody.
- It is possible to spoof this by bypassing the client entirely. See [this Discord conversation](https://discord.com/channels/348658686962696195/423167304956903428/821455192292982805) for more information on this.

I also added a catch in `process_configuration.py` that came up while I was working on this.

I'm sure I forgot something. Tell me if I did and I will fix it.

### DIFF
--- a/src/main/python/rlbot/matchcomms/client.py
+++ b/src/main/python/rlbot/matchcomms/client.py
@@ -68,7 +68,6 @@ class MatchcommsClient:
         self.thread.join(1)
         assert not self.thread.is_alive()
 
-
 async def read_into_queue(websocket: WebSocketClientProtocol, incoming: Queue, team):
     async for message in websocket:
         try:
@@ -91,9 +90,12 @@ async def send_from_queue(websocket: WebSocketClientProtocol, outgoing: Queue, t
             await asyncio.sleep(0.01)
 
         obj = outgoing.get_nowait()
+        # Be verbose if we are overriding the team attr
+        if "team" in obj and obj["team"] != team:
+            print("Overriding the 'team' key in your match communication object!")
         # if we aren't a script, add our team to the packet so that team-only matchcomms can work
-        if team is not None:
-            obj["team"] = team
+        obj["team"] = team
+
         try:
             json_str = json.dumps(obj) # Serialize the object that was put on the outgoing queue
         except TypeError:
--- a/src/main/python/rlbot/matchcomms/client.py
+++ b/src/main/python/rlbot/matchcomms/client.py
@@ -13,7 +13,10 @@ from rlbot.utils.logging_utils import get_logger
 from rlbot.matchcomms.shared import MatchcommsPaths, JSON
 
 class MatchcommsClient:
-    def __init__(self, server_root_url: URL):
+    def __init__(self, server_root_url: URL, team=None):
+        # If self.team is None, you can see all matchcomms
+        self.team = team
+
         self.incoming_broadcast = Queue()
         self.outgoing_broadcast = Queue()
 
@@ -36,8 +39,8 @@ class MatchcommsClient:
             async with websockets.connect(urlunparse(self.broadcast_url)) as websocket:
                 io_task = self.event_loop.create_task(asyncio.wait(
                     [
-                        read_into_queue(websocket, self.incoming_broadcast),
-                        send_from_queue(websocket, self.outgoing_broadcast),
+                        read_into_queue(websocket, self.incoming_broadcast, self.team),
+                        send_from_queue(websocket, self.outgoing_broadcast, self.team),
                     ],
                     return_when=asyncio.FIRST_COMPLETED
                 ))
@@ -66,15 +69,21 @@ class MatchcommsClient:
         assert not self.thread.is_alive()
 
 
-async def read_into_queue(websocket: WebSocketClientProtocol, incoming: Queue):
+async def read_into_queue(websocket: WebSocketClientProtocol, incoming: Queue, team):
     async for message in websocket:
         try:
-            incoming.put(json.loads(message))
+            decoded_message = json.loads(message)
         except json.decoder.JSONDecodeError:
             print('Failed to decode this message:', repr(message))
             traceback.print_exc()
+        # make sure we can't see the message if it is just for the other team
+        if ('team' in decoded_message and 'team_only' in decoded_message and
+                team is not None and decoded_message['team_only'] == True and
+                decoded_message['team'] != team):
+            continue
+        incoming.put(decoded_message)
 
-async def send_from_queue(websocket: WebSocketClientProtocol, outgoing: Queue):
+async def send_from_queue(websocket: WebSocketClientProtocol, outgoing: Queue, team):
     logger = get_logger('matchcomms_json_encode')
     while True:
         # TODO: use something like https://github.com/aio-libs/janus to avoid polling
@@ -82,6 +91,9 @@ async def send_from_queue(websocket: WebSocketClientProtocol, outgoing: Queue):
             await asyncio.sleep(0.01)
 
         obj = outgoing.get_nowait()
+        # if we aren't a script, add our team to the packet so that team-only matchcomms can work
+        if team is not None:
+            obj["team"] = team
         try:
             json_str = json.dumps(obj) # Serialize the object that was put on the outgoing queue
         except TypeError:
--- a/src/main/python/rlbot/agents/base_agent.py
+++ b/src/main/python/rlbot/agents/base_agent.py
@@ -183,7 +183,7 @@ class BaseAgent(RLBotRunnable):
                              "ignore matchcomms in your case then go ahead and wrap your matchcomms access "
                              "in a try-except, or do a check first for whether matchcomms_root is None.")
         if self._matchcomms is None:
-            self._matchcomms = MatchcommsClient(self.matchcomms_root)
+            self._matchcomms = MatchcommsClient(self.matchcomms_root, self.team)
         return self._matchcomms  # note: _matchcomms.close() is called by the bot_manager.
 
     def load_config(self, config_object_header):
--- a/src/main/python/rlbot/utils/process_configuration.py
+++ b/src/main/python/rlbot/utils/process_configuration.py
@@ -99,7 +99,7 @@ def is_process_running(program, scriptname, required_args: Set[str]) -> Tuple[bo
             p = process.name()
             if program in p or scriptname in p:
                 matching_processes.append(process)
-        except psutil.NoSuchProcess:
+        except (psutil.NoSuchProcess, psutil.AccessDenie):
             continue
     # If matching processes were found, check for correct arguments.
     if len(matching_processes) != 0:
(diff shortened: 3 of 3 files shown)


oxrock:
Excited this is happening :)

src/main/python/rlbot/matchcomms/client.py
```diff
@@ -66,22 +69,31 @@ def close(self):
...
     logger = get_logger('matchcomms_json_encode')
     while True:
         # TODO: use something like https://github.com/aio-libs/janus to avoid polling
         while outgoing.empty():
             await asyncio.sleep(0.01)
 
         obj = outgoing.get_nowait()
+        # if we aren't a script, add our team to the packet so that team-only matchcomms can work
+        if team is not None:
+            obj["team"] = team
```
tarehart, line 96: Seems like this makes the 'team' attribute reserved. If somebody already had team in their message, e.g. something like
```
{ name: 'MyBot', team: ['MyAlly1', 'MyAlly2']}
```
then this would overwrite their intended data, right? That might be OK, but we should think of ways to mitigate that problem. Is there a different implementation that could avoid it? If not, are people OK with it? Could we at least warn people if their data is getting overwritten?

VirxEC: good idea. maybe something like:
```python
if obj.get("team"): print("Overriding the 'team' key in your match communication object!")
obj["team"] = team
```

VirxEC:
I just remembered!! Can we also get the index of the bot sending the packet auto-appended, no matter what?

src/main/python/rlbot/matchcomms/client.py
tarehart: Another angle on it: why does this belong in the matchcomms layer rather than the TMCP layer?

ViliamVadocz:
> I just remembered!! Can we also get the index of the bot sending the packet auto-appended, no matter what?

What would you expect for scripts? They don't have an index.

VirxEC:
> What would you expect for scripts? They don't have an index.

Index -1? If it doesn't have an index, then it isn't a standard TMCP packet

oxrock:
> > I just remembered!! Can we also get the index of the bot sending the packet auto-appended, no matter what?
> 
> What would you expect for scripts? They don't have an index.

ignoring tmcp for the moment, this is a good question for scripts in general. Is there a special observer team that can recieve both team's packets? Mayhaps there should be.

ViliamVadocz:
> ignoring tmcp for the moment, this is a good question for scripts in general. Is there a special observer team that can recieve both team's packets? Mayhaps there should be.

According to PR message:
> Scripts can see all matchcomms, even team-only ones, and the ones they send are visible to everybody.

I have a TMCP tracker script that's useful for debugging. Maybe I should put it in the bot pack?

src/main/python/rlbot/matchcomms/client.py
rivques: > why does this belong in the matchcomms layer rather than the TMCP layer?
Because the team of the sender has to be known so the client knows who it came from and if they should be allowed to see it

rivques:
>  Is there a special observer team that can recieve both team's packets? 

Yes. If team is set to `None` or omitted in the `MatchcommsClient` constructor, they can see everything. `BaseScript` does this.

src/main/python/rlbot/matchcomms/client.py
rivques: I like Virx's idea.
Maybe: 
```python
if "team" in obj and obj["team"] != team:
    print("Overriding the 'team' key in your match communication object!")
obj["team"] = team
```

VirxEC: The discourage sending packets as another bot (out of sight, out of mind)

ViliamVadocz: This is why I wanted two matchcomms servers, one per team. It would also restrict the spying and malicious messages.

ViliamVadocz: I actually thought I responded to this PR to this effect but I guess my connection was so bad it failed.

VirxEC: The bot can just bypass the client and send the message to whatever server it wants - just like a bot can send inputs for any other bot.

rivques: > This is why I wanted two matchcomms servers, one per team. It would also restrict the spying and malicious messages.

Two things: One, how do we prevent people from joining (maliciously or accidentally) the wrong server? Two, this seems much more breaking for scripts etc. than simply reserving `team` and telling people if it is being changed. 

I'm not sure there's enough public, un-update-able bots using matchcomms yet for this edge case to have happened, and warning people if it does happen so they can change it seems like enough to me.

ViliamVadocz: Scripts would have two matchcomm clients.

Each client is locked to their server. Afaik, there isn't a way to access another teams server, since it is stored in the botmanager instance which bots don't have access to.

rivques: > Each client is locked to their server. Afaik, there isn't a way to access another teams server, since it is stored in the botmanager instance which bots don't have access to.

Assuming by "since it is stored in the botmanager" you mean "since each server is stored in the botmanager," *how* would each client be locked to its server? What prevents someone from just directly accessing the servers with custom clients? The server is just that - a server. Who can access it is not restricted, and I'm not sure it can be.

ViliamVadocz: If they don't have the port they can't make a custom client

rivques: How do you restrict it so that the server and the authorized clients know the port but not the unauthorized ones?

rivques:
Implemented Virx's verbosity code when overriding the "team" attr

src/main/python/rlbot/matchcomms/client.py
tarehart: > > why does this belong in the matchcomms layer rather than the TMCP layer?
> 
> Because the team of the sender has to be known so the client knows who it came from and if they should be allowed to see it

I don't understand the reasoning. Isn't there a TMCP client in python that could be initialized with the proper team?

rivques: Using fewer pronouns for clarity.

Passing the team in the packet isn't a TMCP problem. (Well, team-only matchcomms was made in part because TMCP wants it, but that's not what this is about.) The problem here is making sure potential receivers of the packet know what team the sender is on so the potential receivers know if they are allowed to look at it. The only way that I have thought of to pass along the team information is by putting it in the packet, since the packet is the only thing that's shared at all.

VirxEC:
Can we get the automatic index appending, too?

ViliamVadocz (commented):
What if the object is not a dictionary? This wouldn't work.

src/main/python/rlbot/matchcomms/client.py
tarehart: I know that you're intending to give the team only functionality to more than just TMCP, I'm questioning whether that's in line with the design philosophy of matchcomms. As I understand it, matchcomms was designed by Dom to be totally unbiased about how it gets used. Generally that's a smart way to avoid regrets later, e.g. if some idea comes along that you didn't imagine at first, and now the system is too specific to be useful and gets replaced.

This is a one-way door, this change cannot be undone. I think team filtering is broadly useful, but I don't feel good about risking the longevity of matchcomms. I recommend either:
- Getting approval from @DomNomNom or,
- Making this a third layer that sits between TCMP and matchcomms.

src/main/python/rlbot/matchcomms/client.py
```diff
@@ -91,9 +90,12 @@ async def send_from_queue(websocket: WebSocketClientProtocol, outgoing: Queue, t
             await asyncio.sleep(0.01)
 
         obj = outgoing.get_nowait()
+        # Be verbose if we are overriding the team attr
+        if "team" in obj and obj["team"] != team:
+            print("Overriding the 'team' key in your match communication object!")
```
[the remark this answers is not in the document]
rivques: Yes, see https://github.com/RLBot/RLBot/blob/master/src/main/python/rlbot/matchcomms/client.py#L74

src/main/python/rlbot/matchcomms/client.py
ViliamVadocz: > I recommend either:
> - Getting approval from DomNomNom or,
> - Making this a third layer that sits between TCMP and matchcomms.

As a third party layer we already have my handler (tmcp package) which filters by team, and also specifies team in messages.

rivques:
How about this:
Each object that goes through the server looks something like this:
```json
{
"origin_team": int,
"origin_index": int,
"team_only": bool,
"matchcomms_obj": array or object
}
```
Where `matchcomms_obj` is the input object, `origin_team` and `origin_index` are -1 if a script, and `team_only` if `true` and sent by a script will only send to other scripts. To implement this, in `MatchcommsClient`, `Queue` will be subclassed to look something like this:
```python
class MatchcommsQueue(Queue):
    def __init__(maxsize=0, index=-1, team=-1):
        self.index = index
        self.team = team
        return super().__init__(maxsize)
    
    def put(item, block=True, timeout=None, team_only=False):
        new_item = {"origin_team": self.team,
            "origin_index": self.index,
            "team_only": team_only,
            "matchcomms_obj": item}
        
        return super().put(new_item, block, timeout)

    def get(block=True, timeout=None, get_metadata=False):
        in_obj = super().get(block, timeout)
        return in_obj if get_metadata else in_obj["matchcomms_obj"]

```
`send_from_queue` and `read_into_queue` will be modified to work with this.

This does not reserve `team`, automatically adds information about the sender as @VirxEC wanted, allows for future expansion, and is backwards-compatible.

src/main/python/rlbot/matchcomms/client.py
VirxEC: The team-only is an extra feature, not mandatory. Setting `team_only` to `True` in the packet will prevent the packet from being sent to the other team. Otherwise, it will be sent to both teams. (At least this was the original idea)

tarehart: To complete the middle layer, you'd need to make a utility that helps you send standard team-stamped (and index-stamped?) messages which is separate from the core matchcomms implementation. Otherwise you're still narrowing matchcomms in an irreversible manner.

rivques:
Closing due to inactivity and abandonment by myself.

PR #552, opened 2021-03-24
Add option to C# bot manager to crash on error

TheBlocks:
This is a quality of life feature for debugging with a debugger. Previously, the debugger would not break whenever a crash occurred.


tarehart:
Is it possible to configure the debugger in Visual Studio to break even on handled exceptions and get the result you want without this code change? https://docs.microsoft.com/en-us/visualstudio/debugger/managing-exceptions-with-the-debugger?view=vs-2019

TheBlocks:
Oh dang, I didn't know about this feature in VS. Seems like Rider also has a similar feature so I'll be closing this PR.

PR #553, opened 2021-03-25
Changes to RLBot's personal virtual environments

VirxEC:
- Requirements won't be checked if the user doesn't have an internet connection (checked with ping to google.com)
- Requirements won't be checked multiple times (e.x. currently in a 3v3 with ABot, RLBot checks the requirements 6 times instead of 1)

This leads to some matches starting faster


tarehart (approved):
Looks good, thank you for making this feature more complete!

PR #555, opened 2021-04-08
Add DFHStadium_Circuit

Darxeal:
Shoutout to tarehart for making possible to add this in the open source part.


VirxEC (approved):
Does support for the map also have to be added into the GUI, or is it automatic? Either way this looks good

PR #556, opened 2021-04-17
Upgrade RLBotDotNet to 2.0.0

TheBlocks:
Breaking changes include:
* Casing of Name, Team, and Index fields. This is so that we can be more consistent with the C# community's guidelines for naming.
* Use of System.Drawing.Color instead of System.Windows.Media.Color. This was a necessary change due to .NET Standard's cross-platform nature.

These changes continue the the changes started in commit https://github.com/RLBot/RLBot/commit/0de12be49c7f9e8c88545b7b248710a7cef18cac.

In terms of NuGet, the nuspec file was deleted in favour of embedded NuGet information directly into the csproj file. I made this change because Microsoft seems to heavily recommend this way instead of having a nuspec for projects targeting newer versions of .NET.

I've also deleted unnecessary files from the TestBot like AssemblyInfo.cs, App.config, and packages.config (no longer required because of .NET Standard).

The release notes for the NuGet package link to https://github.com/RLBot/RLBotCSharpExample/wiki/Upgrading-to-RLBotDotNet-2.x so that users can fix any errors due to breaking changes.


NicEastvillage (approved):
Good job with the updating guide!

VirxEC (approved):
Looks fine to me, Red also said that he approves. Merging.

PR #567, opened 2021-11-07
Updated to flatbuffers 2.0

ItsCodeRed:
Updated to work with the latest version of flatbuffers.

Note that it seems that flatbuffers isn't routinely checked and updated by the rlbotgui, so this change could potentially cause issues for people without the latest version of flatbuffers. Of course, they can just update the package themselves, but I think it would be nice if the gui did it for them. I would go and make this change myself, but I don't really know how I would do that.


tarehart:
Thanks for the PR!

Upgrading flatbuffers will be a little intense, because lots of binaries are currently running version 1.9:
- RLBot.exe
- Every Java bot
- Every C# bot
- Every C++ bot

I don't know whether flatbuffers has full binary compatibility between 1.9 and 2.0, and we'd need to test carefully even if they claim that. It might be worth doing, but first I'm going to focus on getting 1.9 working smoothly for  people again.

tarehart:
I've changed it so we're just pinning to 1.9 as Virx suggested, look good to you? If so, we can squash merge it with an appropriate commit message.

ItsCodeRed:
Looks good!

ItsCodeRed:
I’m still pretty new to using git, is the squash merge something I have to do?

tarehart:
Nah I can take care of it

tarehart (approved):

PR #569, opened 2021-12-06
Fix for Rocket League startup on Epic Games Launcher.

tarehart:
Our `webbrowser.open( 'com.epicgames.launcher://apps/Sugar?action=launch&silent=true')` technique isn't working anymore, but this new code works for me.

Hopefully fixes https://github.com/RLBot/RLBotGUI/issues/174

I'm also making a small change to the way we kill processes, sending SIGTERM ahead of time so they have a chance to clean up.

### DIFF
--- a/src/main/python/rlbot/gamelaunch/epic_launch.py
+++ b/src/main/python/rlbot/gamelaunch/epic_launch.py
@@ -1,3 +1,4 @@
+import psutil
 import time
 
 import re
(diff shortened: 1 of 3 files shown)


VirxEC:
Looks good to me, my one gripe being that we have a time limit at all. Seems to only cause issues, but can be resolved in another pr

tarehart:
Good point, I'm killing the time limit.

PR #570, opened 2021-12-10
fixed unhandled exception when doing epic startup trick

SavageSnowgoose:
psutil.NoSuchProcess was not handled for one code path during launch_with_epic_login_trick which would lead to it falling back to trying to start the old way, which means it would just end up with a RL without the rlbot flags.

### DIFF
--- a/src/main/python/rlbot/utils/process_configuration.py
+++ b/src/main/python/rlbot/utils/process_configuration.py
@@ -116,8 +116,12 @@ def is_process_running(program, scriptname, required_args: Set[str]) -> Tuple[bo
                     return True, process
             except psutil.AccessDenied:
                 print(f"Access denied when trying to look at cmdline of {process}!")
-        # If we didn't return yet it means all matching programs were skipped.
-        raise WrongProcessArgs(f"{program} is not running with required arguments: {required_args}!")
+            except psutil.NoSuchProcess:
+                # process died whilst we were looking at it, so pretend we found none, it will get handles another time
+                break
+        else:
+            # If we didn't return yet it means all matching programs were skipped.
+            raise WrongProcessArgs(f"{program} is not running with required arguments: {required_args}!")
     # No matching processes.
     return False, None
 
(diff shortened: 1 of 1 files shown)


src/main/python/rlbot/utils/process_configuration.py
```diff
@@ -116,8 +116,12 @@ def is_process_running(program, scriptname, required_args: Set[str]) -> Tuple[bo
                     return True, process
             except psutil.AccessDenied:
                 print(f"Access denied when trying to look at cmdline of {process}!")
-        # If we didn't return yet it means all matching programs were skipped.
-        raise WrongProcessArgs(f"{program} is not running with required arguments: {required_args}!")
+            except psutil.NoSuchProcess:
+                # process died whilst we were looking at it, so pretend we found none, it will get handles another time
+                break
```
tarehart, line 121: The other except block doesn't use break, it allows the loop through matching processes to continue. Is there a good reason to break here? For epic launch it'll probably be a list of one and not matter, but this fn might get used for other purposes also.

tarehart (commented):
I'm extremely happy to get bug fix contributions, thank you SavageSnowgoose!

src/main/python/rlbot/utils/process_configuration.py
SavageSnowgoose: By breaking here we get the same outcome to the function as if we had hit the same  race condition when checking the process name at line 102.

SavageSnowgoose: I have pushed a better fix. The open question I have is whether the access denied should lead to the return False path or the WrongProcessArgs path. My thoughts are that if we have permission denied then it probably implies the process in question is being run by another user and as a result we should pretend it doesn't exist rather than it existing but being incorrect. Also the exception implies that the args for the found process are wrong and we cannot know that if we got access denied.

tarehart (approved):
This update looks great, thanks!

src/main/python/rlbot/utils/process_configuration.py
tarehart: Yep I agree.

PR #574, opened 2022-01-25
Fix for "Trouble with epic launch: 'NoneType' object has no attribute 'kill'" error

SavageSnowgoose:
Error can occur if the process dies between when we first saw it and grabbed the args, and then when we go to kill it where we were looking for the process via iterating all processes again and assuming success.

Fix removes the second iterating, as to reach this point we have already succeeded. If the process has already been killed we will no longer error as we can call `kill` on an already dead process without error. 
In the case it did die to have reached this point we did see it and did get the args, so no reason not to push ahead regardless.

Also includes a refactor - the is_process_running returned a bool and a process/None. It was being used in a way where the bool would be checked and then assume the process is not None based on the bool. If the bool and process is always guaranteed to be tied why return both. If there is a case where one can envisage the bool and process not being coupled in that way then existing use cases will break. The refactor removes the chance for the ambiguity. Refactor has a new function name, old one is still kept for backwards compatibility.

### DIFF
--- a/src/main/python/rlbot/gamelaunch/epic_launch.py
+++ b/src/main/python/rlbot/gamelaunch/epic_launch.py
@@ -109,7 +109,6 @@ def launch_with_epic_login_trick(ideal_args: List[str]) -> bool:
             time.sleep(1)
             logger.info("Waiting for Rocket League args...")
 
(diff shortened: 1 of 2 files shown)


tarehart (approved):
Looks good, thanks for keeping backwards compatibility!

PR #575, opened 2022-01-30
Fixed bug in last commit

SavageSnowgoose:
Last commit not working for some, the process variable can be holding the original or restarted RL exe.

This now reverts that assumption. Instead just targets the issue the user was having with 1.62.3 where the process did not exist (so it was trying to call 'kill' on None). Instead we check that it is not None, and then also handle the case that the process may die before we kill it.

Still unsure why the RL would have dies for that user, (perhaps they had killed it?) but this fix still should give the intended outcome of the last fix: we have the args, the fact the process has dies outside of our control is OK as we have enough to continue.


tarehart (approved):
Looks good! Thanks for following up on this and adding such good commentary.

PR #577, opened 2022-02-06
Avoiding ghost inputs and failed bot starts when using training.

tarehart:
The main part of this change is to take more care with `agent_metadata_queue`. Messages in that queue can start bots if they contain a helper process request, so it's very bad if messages from a previous match carry over to the next match somehow. Even if the bots from the previous match were properly shut down or killed, they'll be started again and cause "ghost inputs" where more than one bot process is trying to control the same bot index.

This was impacting non-python bots that rely on the helper process system, e.g. Scratch, C#, Java (and of course any bot can be impacted by the subsequent ghost inputs).

While I was in there, I made some other changes to make the training system work better for the purposes of AutoLeague type programs.


NicEastvillage (approved):
Nice exorcism there.

PR #579, opened 2022-02-12
Fix Linux launch

VirxEC:
If someone could please test this in Windows for me that would be awesome haha, just start a match and if it works then it should be good. I just want to make sure that I didn't break anything for you Windows folk.

I haven't fixed anything with Proton, just fixed the breaking changes introduced in 1.53.0.


VirxEC:
Confirmed, no bugs appear to have been introduced into the Windows launch with these changes.

tarehart (approved):
Just when I think I'm done with linux they pull me back in :)

Thanks for restoring this functionality.

PR #580, opened 2022-02-17
Fix RLBot's virtual environments

VirxEC:
Changes:

- clear by default is False
- Set with_pip to true (auto-install pip and setuptools for us)
- Use `python -m pip -U pip` to upgrade pip
- Make the create method configure using system site-packages before calling `post_setup`

### DIFF
--- a/src/main/python/rlbot/utils/virtual_environment_management.py
+++ b/src/main/python/rlbot/utils/virtual_environment_management.py
@@ -1,7 +1,7 @@
+import os
 import sys
 from pathlib import Path
-from subprocess import PIPE, Popen, run
-from threading import Thread
+from subprocess import run
 from types import SimpleNamespace
 from typing import List
 from venv import EnvBuilder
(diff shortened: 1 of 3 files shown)


src/main/python/rlbot/utils/virtual_environment_management.py
```diff
@@ -28,8 +59,9 @@ def post_setup(self, context: SimpleNamespace) -> None:
         sys.stderr.write(f'Installing {requirements}...\n')
         sys.stderr.flush()
 
-        args = [context.env_exe, '-m', 'ensurepip']
-        finished_process = self.run_and_dump(args, timeout=120)
+        # Actually update pip
+        args = [context.env_exe, '-m', 'pip', 'install', '-U', 'pip']
+        finished_process = self.run_and_dump(args, timeout=300)
```
tarehart, line 64: What's the story on this? Upgrading pip seems like a good idea, but is it safe to remove `env_exe -m ensurepip`? Are we guaranteed to always already have pip at this point?

src/main/python/rlbot/utils/virtual_environment_management.py
```diff
@@ -12,10 +12,41 @@
...
+        Create a virtual environment in a directory.
+        Changed so system site packages is configured before post_setup is ran.
+
+        :param env_dir: The target directory to create an environment in.
+        """
+
+        env_dir = os.path.abspath(env_dir)
+        context = self.ensure_directories(env_dir)
+        # See issue 24875. We need system_site_packages to be False
+        # until after pip is installed.
```
tarehart, line 33: Interesting! Looks like the referenced issue is https://bugs.python.org/issue24875. It's old and allegedly resolved, do we still need this code?

src/main/python/rlbot/utils/virtual_environment_management.py
VirxEC: Yes, with_pip guarantees that we will.

src/main/python/rlbot/utils/virtual_environment_management.py
VirxEC: That comment is actually from the original code in Python, I just copy+pasted the function and made post_setup() get called last. https://github.com/python/cpython/blob/3.7/Lib/venv/__init__.py#L52-L76

tarehart (approved):
Looks good!

PR #582, opened 2022-03-17
Epic launch fix

VirxEC:
When searching for the exe in Windows, check if the found path is actually a valid file before returning. It seems that users that have moved their Rocket League installs between drives have been having issues with this. Hopefully, this fixes that!

[Example of the issue on Discord](https://discord.com/channels/348658686962696195/348659150793736193/953972576365998090)

Image:
image


tarehart (approved):
Good idea!

PR #584, opened 2022-03-25
Fix launching bots with venvs on non-windows

VirxEC:
If a bot utilized a virtual environment, then it just wouldn't launch if the user wasn't on Windows. This, at least, should work on Windows.


NicEastvillage (approved):
I assume you know what you are doing

VirxEC:
I tested it and it worked, so I'm about as sure as I can hope for

PR #586, opened 2022-05-07
Add a launcher preference for epic_only, without steam fallback

tarehart:
Currently we accept launcher preferences for steam vs epic. The steam choice tries steam only. The epic choice tries epic, and then falls back to steam if that fails, which causes a large amount of confusion and misery for our users.

Adding a new option (for backwards compatibility) called epic_only that does NOT fall back to steam. A separate change to RLBotGUI can rename the UI options to match the actual behaviors.

Also adding an input that RLBotGUI could use later to set the location of the Epic RocketLeague.exe file.

### DIFF
--- a/src/main/python/rlbot/gamelaunch/epic_launch.py
+++ b/src/main/python/rlbot/gamelaunch/epic_launch.py
@@ -15,11 +15,16 @@ from rlbot.utils.logging_utils import get_logger, DEFAULT_LOGGER
 from rlbot.utils.process_configuration import get_process
 
 
-def launch_with_epic_simple(ideal_args: List[str]) -> bool:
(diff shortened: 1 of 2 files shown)


L0laapk3:
Why add a third option for this, rather than a checkbox to allow falling back on the other or something?
Just writing down my thoughts, but it doesnt seem like the route that makes the most sense to me, as well as not covering the case where steam is prefered but both are accepted

tarehart:
Yeah that makes sense. We could go that route, but it would be a little annoying because:
- To make the change backwards compatible / not alter anybody's RLBotGUI behavior, we would need to default the "fallback" checkbox to True, which is extra migration effort.
- Cross-launcher fallback is not a very wise behavior, and I don't want to add more code to allow fallback in the other direction.

PR #594, opened 2022-08-16
Add open_csv for loading the game items

VirxEC:
This is for RLBot/RLBotPack#235

### DIFF
--- a/src/main/python/rlbot/csv/items.py
+++ b/src/main/python/rlbot/csv/items.py
@@ -2,6 +2,6 @@ import os
 from pathlib import Path
 
 
-def open_csv():
-    """Opens RLBot's items.csv file for reading"""
-    return open(os.path.join(Path(os.path.realpath(__file__)).parent, "items.csv"), "r")
+def get_items_path() -> Path:
+    """Returns the path to RLBot's items.csv file for reading"""
+    return Path(os.path.join(Path(os.path.realpath(__file__)).parent, "items.csv"))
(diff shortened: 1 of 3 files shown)


src/main/python/rlbot/csv/items.py
```diff
@@ -0,0 +1,7 @@
+import os
+from pathlib import Path
+
+
+def open_csv():
+    """Opens RLBot's items.csv file for reading"""
+    return open(os.path.join(Path(os.path.realpath(__file__)).parent, "items.csv"), "r")
```
tarehart, line 7: Many people consider it a best practice to use a `with` statement when reading from a file: https://stackoverflow.com/questions/14776853/is-it-good-practice-to-depend-on-pythons-with-as-statement

Are you intending for people to do something like this?
```python
with open_csv() as csv:
   ...
```

That's not very obvious to me, might be good to add a type annotation and a comment, or just return a Path object instead. A Path might be more broadly useful.

tarehart (commented):
Thanks for doing this!

You'll probably need to add `**/*.csv` to the list at https://github.com/RLBot/RLBot/blob/master/src/main/python/setup.py#L39-L48 in order to distribute this csv file through pypi successfully.

src/main/python/rlbot/csv/items.py
VirxEC: Sure, I'll make the change.

Also, yes that was how I hoped they would use it: https://github.com/RLBot/RLBotPack/blob/10d0cc969a91087c8d874548b86f3ead1d979c9b/RLBotPack/Kamael_family/nebulous_loadout.py#L11

VirxEC:
Made the requested changes to both PRs.

tarehart (approved):

PR #595, opened 2022-08-19
Implement fixes for standalone bots

VirxEC:
Every time a bot is launched, it will try to recreate the `venv` folder. If the folder exists, the program will only ensure it's valid and check for updates. This prevents an error that can happen when another program uses the virtual environment and RLBot tries to recreate it, throwing an error saying that `python.exe` is busy. Specifically prevented by avoiding calling the `setup_python` function in `builder.create`. Implemented for both bots and scripts.

Next, I fixed an error where RLBot could forever have `has_received_metadata_from_all_bots` return `False` because it thinks it hasn't gotten the metadata from the standalone bots. In reality, it got it before any of the normal bots, but `num_metadata_received` wasn't incremented which is the root of the issue. Incrementing that value after appending `agent_metadata_map` when starting standalone bots fixes this issue. THIS ISN'T AN ISSUE FOR THE CURRENT GUI, but it is for the Rust GUI as it gets caught in a loop of forever waiting for the metadata.

ADDITIONALLY: Only search for `RocketLeague` instead of `RocketLeague.exe` on non-Windows.

### DIFF
--- a/src/main/python/rlbot/setup_manager.py
+++ b/src/main/python/rlbot/setup_manager.py
@@ -352,7 +352,14 @@ class SetupManager:
                         checked_environment_requirements.add(bundle.requirements_file)
 
                 builder = EnvBuilderWithRequirements(bundle=bundle, do_post_setup=do_post_setup)
-                builder.create(Path(bundle.config_directory) / 'venv')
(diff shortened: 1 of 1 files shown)


ddthj (approved):
Works on my machine!

PR #600, opened 2022-10-15
Add 6 new maps

Darxeal:


VirxEC (approved):
Looks good, sad that this repo doesn't have auto-deployment so we need to wait for Tare or maybe Goose

PR #601, opened 2022-10-30
Make sure stdin is never inherited when spawning subprocess

VirxEC:
This fixes an error that I was encountering in my testing.


skyborgff:
What was the error you were getting?

VirxEC:
Standalone bots wouldn't launch when the parent process was reading stdin, A bug on Windows in the Rust GUI. Found this solution after asking in Stack Overflow.

ddthj (approved):
"and it won't effect any existing code in any way" - Virx

PR #619, opened 2023-07-01
Fix deprecated usage of selenium library in scratch manager

jboardman367:
- Updates usage of selenium library to work with selenium 4
- Require `selenium==4.*`


NicEastvillage (approved):
I tried to test it. Not sure if I did it right, but I got the Scratch editor to open. Lets try.

